PR #103, opened 2019-03-07
Modified continuum.adoc making incremental changes

jclingan:
Title is self-explanatory :-)


vorburger:
@jclingan based on https://github.com/quarkusio/quarkusio.github.io/pull/118 I think you may need to change this at the source over in quarkusio/quarkus instead of here, see https://github.com/quarkusio/quarkusio.github.io/pull/121.

cescoffier:
So, it's where it's confusing....

Topics are on the web site 
Guides on on the main repo.

As I said on another issue, we are discussing a better workflow.

PR #114, opened 2019-03-07
Remove a remaining reference to old project name

gsmet:


kenfinnigan:
Not sure that file is actually surfaced on the website, it uses the index file instead

vorburger:
@gsmet based on https://github.com/quarkusio/quarkusio.github.io/pull/118 I think you may need to change this at the source over in quarkusio/quarkus instead of here, see https://github.com/quarkusio/quarkusio.github.io/pull/121.

PR #118, opened 2019-03-07
remove Maven plugin from Gradle example

vorburger:
The Gradle maven plugin is only used in projects which need to re-use Gradle built artifacts in Maven projects. This will not be the case in a typical Gradle built Quarkus project.

PR #121, opened 2019-03-07
update README clarifying that MD files in this repo are all generated

vorburger:
based on #118


vorburger:
@cescoffier

cescoffier:
It's only for guides.... Topics are stored in this repository. Yes... confusing... Even for us.

PR #125, opened 2019-03-08
Sync the changes made by Gunnar to the transaction guide

gsmet:
@cescoffier can you review and merge, please?

Better have these fixes as soon as we can.


cescoffier:
Done! Thanks for the head's up.

PR #127, opened 2019-03-08
change “you” to “your”

darinpope:


cescoffier:
Thanks! I've applied your fix to a PR on the quarkus repository (which is the "true" source for the guides).

Issue #110, opened 2019-03-07
Characters like p, q, j, g are cut at the bottom in code and command examples


rsvoboda:
Characters like p, q, j, g are cut at the bottom in code and command examples.

<img width="540" alt="screenshot 2019-03-07 at 10 40 53" src="https://user-images.githubusercontent.com/925259/53947817-d0112e80-40c6-11e9-9346-1069ff9510ca.png">
<img width="540" alt="screenshot 2019-03-07 at 10 41 11" src="https://user-images.githubusercontent.com/925259/53947818-d0112e80-40c6-11e9-9a01-19ee3b72ca89.png">

I tried to fix it via https://github.com/rsvoboda/quarkusio.github.io/commit/b4e22b1ef2feb44e12989ac07ffae5c42e9e564e but it has side effects :(

gekorob:
Hi all,
  is it possible to add a *line-height: 1.4;* to the *.hljs* class?

```scss
.hljs {
  display: inline;
  overflow: inherit;
  overflow-x: inherit;
  overflow-y: hidden;
  background: inherit;
  padding: 0;
  line-height: 1.4;
}
``` 
in the *quarkus.scss* ?

This shoud fix the problem for the highlights section.

rsvoboda:
Nice, `.hljs` css update fixes the issue.

@gekorob, do you want to create PR or should I do that instead of you ?

gekorob:
Yes, I'd like to create the pull request if for you it's ok.

PR #131, opened 2019-03-09
Characters like p, q, j, g are cut at the bottom in code and command examples #110

gekorob:
This fix tries to address the problem of the of line height in highlight boxes.


rsvoboda:
This should fix #110

PR #133, opened 2019-03-10
Added patch command to specify Dockerfile location

sabre1041:
Specifying the location of the Dockerfile used to produce the image


cescoffier:
@sabre1041 wow... this is a great trick! Thanks for that. 
The guide source is on the main quarkus repository, but don't worry, I'm updating it right now!

PR #135, opened 2019-03-10
Improve the Gradle tooling documentation

jnizet:
- Provide Gradle Kotlin DSL examples
- Import the Quarkus BOM to add dependencies without specifying versions (as done with Maven)
- Avoid applying the plugins imperatively, as recommended, for a better integration of the Gradle Kotlin DSL
- Use the project settings to download the plugin, instead of modifying the global Gradle settings
- Use the Gradle wrapper in the shell examples
- Fix a typo in the Eclipse integration

See https://github.com/jnizet/quarkus-modern-gradle-example for a project using the Gradle code suggested in this commit (the kotlin-dsl branch contains the example using the Kotlin DSL)

### DIFF
--- a/_guides/gradle-config.adoc
+++ b/_guides/gradle-config.adoc
@@ -3,26 +3,42 @@ include::./attributes.adoc[]
 
 // tag::repositories[]
 {project-name} Gradle plugin is not yet published to the https://plugins.gradle.org/[Gradle Plugin Portal],
-so you need to add the following to your '~/.gradle/init.gradle' file:
+so you need to add the following at the top of your './settings.gradle' file:
 [source, groovy, subs=attributes+]
 ----
-allprojects {
-  buildscript { <1>
+pluginManagement {
     repositories {
-      mavenCentral()
+        mavenCentral()
+        gradlePluginPortal()
     }
-    dependencies {
-      classpath 'io.quarkus:quarkus-gradle-plugin:{quarkus-version}'
+    resolutionStrategy {
+        eachPlugin {
+            if (requested.id.id == 'io.quarkus.gradle.plugin') {
+                useModule("io.quarkus:quarkus-gradle-plugin:${requested.version}")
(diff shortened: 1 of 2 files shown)


geoand:
Thanks for the PR @jnizet!

One thing that I could not get to work at all was the "BOM" resolution. Basically our build process does some heavy lifting with the dependencies in order to create the various jars and it seems like the BOM declaration breaks them.

jnizet:
Thanks for the feedback @geoand. 

I included it because I felt it was weird to have a BOM and use it only in Maven, although Gradle supports BOMs too.
If I understand your comment correctly, you're saying that the Quarkus gradle plugin doesn't deal with dependencies correctly when their version number is resolved thanks to the BOM. I'd say it's something that should be fixed in the plugin, but I understand that it's a harder task than just changing the documentation, so maybe I can just amend this PR so that it avoids importing the BOM (until it's supported), and hard-codes version numbers as it was done before?

Is that alright with you?

An issue should probably be opened in order to properly support using the BOM, but I'll let you do it, because I don't know exactly what doesn't work.

geoand:
@jnizet Yeah, I would say for now best to amend to PR to include the versions despite BOM being present.

Also I am definitely no expert in Gradle so I don't really know why the dependency resolution is failing, but I will open an issue and see what we can figure out.

Also the PR will definitely need more eyes from people more experienced in Gradle than myself :)

gsmet:
@jnizet sorry about the confusion but for now, the guides related PRs have to be submitted to the main repo (see the `docs/src/main/asciidoc/` directory).

Could you resubmit your PR for the main `quarkus` repo?

Thanks!

jnizet:
@gsmet I will, but could you please fix the CONTRIBUTING.md file then, which I read to know how to contribute, and which doesn't provide that information at all, and instead asks for PRs here against the develop branch?

PR #136, opened 2019-03-11
Fix typo in user home dir syntax

CedricGatay:
There is only a few chances that the current person reading the guide unpack the GraalVM distribution into a home directory for a user named `clement` 😉. 

Fixed this by removing the user name in the syntax and explicitely tell the user to expand the archive into its home directory.


cescoffier:
The issue has been fixed and the web site updated.

Thanks for reporting!

PR #148, opened 2019-03-24
Remove extra 'l' in microprofile.config values

HasseNasse:


gsmet:
Thanks, AFAICS it has already been fixed in the main repo. The documentation will be deployed tomorrow when we release the new version.

PR #151, opened 2019-03-26
fix(faq): Use correct closing bracket in hyperlink

dsimansk:
The hyperlink is not rendered correctly due to incorrect closing bracket.


gsmet:
Good catch, thanks!

PR #152, opened 2019-03-27
rename gradle plugin name

literalice:


gsmet:
Thanks, this has already been fixed in the main repo. It will deployed tomorrow when we release the new version officially.

PR #153, opened 2019-03-29
fixes for container runtime options on building native image on container

literalice:
It's for container runtime options for building a native image( quarkusio/quarkus#1722 ).


literalice:
@starksm64 Thank you for the review. I applied the change on quarkusio/quarkus#1722 PR.

PR #166, opened 2019-04-10
List the Swagger UI extension

danielpetisme:
Add Swagger UI to the extensions list.

### DIFF
--- a/_data/extensions.yaml
+++ b/_data/extensions.yaml
@@ -62,6 +62,13 @@ categories:
         - open-api
         groupId: io.quarkus
         artifactId: quarkus-smallrye-openapi
+      - name: Swagger UI
+        description: "Generate a Swagger UI from your OpenAPI specification"
+        labels:
+        - swagger-ui
+        - swaggerui
+        groupId: io.quarkus
+        artifactId: quarkus-swagger-ui
       - name: SmallRye REST Client
         description: "Call REST services"
         labels:


emmanuelbernard:
> TBH, I didn't add it here on purpose as I wasn't sure we should advertise it outside of the OpenAPI extension.
> I would maybe rather add "Comes with Swagger UI" in the OpenAPI description.

so the JSON drives the list-extensions mode. Is a user expected to add swagger ui in his pom? Then I think it would be in. But on the guide side, it should all be in the OpenAPI description. That's my current decision framework. For example some extensions are only dependencies of other extensions right now, and not exposed to the user, for these we don't list them in the JSON and extension page.

gsmet:
Currently, the situation is:
- Swagger UI is in `extensions.json`
- it's not on the website

This extension is a bit special because you could use the Swagger UI extension alone and point it to a generated OpenAPI descriptor available somewhere else but keep in mind it's only available in dev mode.

The Swagger UI extension is a dependency of the OpenAPI extension and most users should use it this way, even if they can use it to point to other locations.

I think I would remove it altogether and make it clear on the website that OpenAPI comes with Swagger UI. Advanced users can install the extension if they want but it's advertised as a component of the OpenAPI one.

As for the guide, it's one an only guide.

gsmet:
I added Swagger UI to the OpenAPI extension (and added a bunch of other extensions that were missing)

PR #171, opened 2019-04-11
Fix typo

psycho-ir:


gsmet:
Thanks!

For the record, the fixes to the guides should be made in the main repo (I added a header in the files but apparently it's not sufficient :)).

I will merge this one and report it in the main repo so that it doesn't get lost.

Thanks!

psycho-ir:
@gsmet Oh I see, I think the reason is that those comments are just visible in raw mode so people can easily overlook them. 
I just opened this #175, hopefully it helps. wdyt?

PR #180, opened 2019-04-24
Panache guide: specify how to run write operations

thescouser89:
With this commit, the Panache guide specifies that we should run the
`persist` operations inside a CDI method annotated with `@Transactional`

### DIFF
--- a/_guides/hibernate-orm-panache-guide.adoc
+++ b/_guides/hibernate-orm-panache-guide.adoc
@@ -133,7 +133,8 @@ This allows for proper encapsulation at runtime as all fields calls will be repl
 
 == Most useful operations
 
-Once you have written your entity, here are the most common operations you will be able to do:
+Once you have written your entity, here are the most common operations you will be able to do (run these inside a CDI method
+annotated with `@Transactional`):
 
 [source,java]
 --


gsmet:
@thescouser89 the source of the guides are maintained in the main repo here: https://github.com/quarkusio/quarkus/blob/master/docs/src/main/asciidoc/hibernate-orm-panache-guide.adoc .

That being said, I'm not totally sure it should be introduced this way but I agree we need to be clearer about it.

@FroMage can you follow up on this?

thescouser89:
@gsmet Thanks for looking into the PR!

Ah so should I update it on the quarkus repository then? I guess it's on quarkus and then it gets copied to quarkusio.github.io?

And yeah I think it was a bit awkward to introduce it this way but can't quite figure out how to 'say' it better. I'm more than happy to get input on it! :)

FroMage:
Mmm, thanks. I think I would have added a "note" element about this somewhere more prominent. I guess the Hibernate ORM guide would have the same issue, I notice that it's only mentioned in passing at https://quarkus.io/guides/hibernate-orm-guide.

I've filed it at https://github.com/quarkusio/quarkus/issues/2266, and will close this one. Thanks a lot.

PR #195, opened 2019-05-13
Make CDI constructor injection example compilable

stil4m:
Set field to `null` in default constructor, while it will not compile otherwise due to the `final` keyword.


mkouba:
I'm closing this PR - it's out of date. The field  is not final anymore and even the no-args constructor is not needed anymore. Anyway, thanks for contribution and for the next time pls send a PR direcly to https://github.com/quarkusio/quarkus where the source of the guide is located: https://github.com/quarkusio/quarkus/blob/master/docs/src/main/asciidoc/cdi-reference.adoc.

PR #202, opened 2019-05-15
remove {project-name} template from page header

jmtd:
{project-name} is not being substituted by the build/publish process, so it's appearing live at https://quarkus.io/guides/ap4k

Looking at the other guides, they typically don't have {project-name} in their heading (any more) so just remove it to be consistent


gsmet:
This needs to be applied in the main repo (`docs` module) otherwise it will be overwritten at the next release.

PR #208, opened 2019-05-17
Added RSS feed

adelasofia:
@emmanuelbernard / @kenfinnigan:

RE: https://github.com/quarkusio/quarkusio.github.io/issues/191#event-2328697691

Added the RSS feed for the site. I've never used an RSS feed myself so it's possible I may have missed something (should it be linked somewhere?). You can see the generated feed on "/feed.xml".

I also set the limit of posts shown in the feed to 10, but this can easily be changed on line # 13 of the feed.xml file in the root directory.


emmanuelbernard:
Thanks @adelasofia I tested it successfully. I have two follow up but will create an issue for that.

PR #210, opened 2019-05-24
Blog post about Angular and Quarkus development.

kabir:

PR #223, opened 2019-06-06
Blog post about Angular and Quarkus development.

kabir:
This replaces https://github.com/quarkusio/quarkusio.github.io/pull/210 and is hopefully simpler and gives us a more positive spin on it.

I spoke to Stuart about the Quarkus router API that Emmanuel mentioned in the old PR, and he recommended not to use it.

### DIFF
--- a/_posts/2019-06-05-quarkus-and-web-ui-development-mode.adoc
+++ b/_posts/2019-06-05-quarkus-and-web-ui-development-mode.adoc
@@ -10,7 +10,7 @@ author: kkhan
 
 In this blog post we will take advantage of the respective development modes of both Quarkus and Angular CLI and see how we can develop a zero turnaround web application backed by a RESTful API on Quarkus. While I am using Angular, *the concepts are the same for other modern web application frameworks.*
 
-The source code for this blog can be found at https://github.com/kabir/blog-quarkus-ui-development. It also contains a README which explains in detail how the application was set up and how we implemented each step, addressing problems we uncovered along our way.
+The source code for this blog can be found at https://github.com/kabir/blog-quarkus-ui-development. It contains a README which explains in detail how the application was set up and how we implemented each step, addressing problems we uncovered along our way.
 
 You need to have https://nodejs.org/en/download/package-manager/[Node], https://yarnpkg.com/en/docs/install[Yarn] and https://cli.angular.io[Angular CLI] installed on your system, and familiarity with Angular and Quarkus is assumed.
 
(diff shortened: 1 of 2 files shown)


rsvoboda:
@emmanuelbernard new version of Kabir's blog post

_posts/2019-06-05-quarkus-and-web-ui-development-mode.adoc
```diff
@@ -0,0 +1,110 @@
...
+The example project has already been set up with the needed adjustments to do this. These are the main things which need doing:
+
+. Make sure we handle the Angular routes in the UI rather than on the back-end. This is important for the bundled version (running on port 8080).
+. Enable the Angular proxy
+.. Signal to the back-end that the front-end is running in a proxy, i.e. not served by us. This is important for when the back-end needs to redirect to a page in the front-end.
+
+We will not go into details of how the application is set up, but will instead focus on the changes needed to make this work well.
+
+==== Handle Angular Routes
+We need to add a filter, https://github.com/kabir/blog-quarkus-ui-development/blob/master/src/main/java/org/kabir/quarkus/ui/AngularRouteFilter.java[AngularRouteFilter]. The essence of its `doFilter()` method is:
```
emmanuelbernard: This is minor but I prefer to replace the We need to sentences with just action sentences.
"We need to add" -> "Add"

_posts/2019-06-05-quarkus-and-web-ui-development-mode.adoc
```diff
@@ -0,0 +1,110 @@
...
+            request.getContextPath().length()).replaceAll("[/]+$", "");
+    if (!FILE_NAME_PATTERN.matcher(path).matches()) {
+        // We could not find the resource, i.e. it is not anything known to the server (i.e. it is not a REST
+        // endpoint or a servlet), and does not look like a file so try handling it in the front-end routes.
+        request.getRequestDispatcher("/").forward(request, response);
+    }
+}
+
+----
+Without this filter, if you try to go to one of the Angular routes directly (e.g if you try to refresh when on http://localhost:8080/rest which is the `Rest` page we saw earlier) Quarkus will think it is a file, a REST endpoint or a servlet. Since it has no knowledge of anything called `/rest` you end up with a 404 (Not Found) status. `AngularRouteFilter` checks the status code, and if it is 404 and does not seem to be a file, we attempt to handle it in Angular.
```
emmanuelbernard: I understood that this filter will redirect localhost:8080/foo/bar to localhost:4200/foo/bar but either ti's not what it is doing precisely or I don't see how that redirection is configured.
Or I did not get the usefulness of the filter.

_posts/2019-06-05-quarkus-and-web-ui-development-mode.adoc
```diff
@@ -0,0 +1,110 @@
...
+
+
+==== Set up the Angular proxy
+We configure the proxy in https://github.com/kabir/blog-quarkus-ui-development/blob/master/webapp/proxy.conf.json[proxy.conf.json]. All REST calls to anything under `/api` will be passed to the back-end running on port 8080. To run the Angular development server with this configuration, we have added a `proxy` configuration to the `scripts` section of https://github.com/kabir/blog-quarkus-ui-development/blob/master/webapp/package.json[`package.json`].
+
+In our case, we have a servlet which needs to redirect back to the front-end (something I found I needed to implement OAuth in my main project). That has a check for the `-Dui.proxy` system property we saw earlier when handling the `/callback` path in https://github.com/kabir/blog-quarkus-ui-development/blob/master/src/main/java/org/kabir/quarkus/ui/SampleServlet.java[`SampleServlet`]. If this property is set, we prepend `https://localhost:4200` (the address of the Angular proxy) to the redirect URL if we find the proxy is running on port 4200.
```
emmanuelbernard: Any reason you don't read the ui.* properties via the `@Inject @ConfigProperty`? That way you coudl read them via applciation.properties, system properties and enviornment properties equally

_posts/2019-06-05-quarkus-and-web-ui-development-mode.adoc
kabir: ok

_posts/2019-06-05-quarkus-and-web-ui-development-mode.adoc
emmanuelbernard: Also I did not understnand why this servlet was only answering to `/servlet/` and not everything. As you can see I'm still a bit lost on the app flow exchange.

emmanuelbernard:
Hey @kabir I made some comments but I don't want to block you on it if you feel you've spent enough time on it. So it's up to you if you want to refine another time to just go for it.

_posts/2019-06-05-quarkus-and-web-ui-development-mode.adoc
kabir: Not really, just to keep it simple (as well as being slightly ignorant of MP config). We only ever want this to be triggered via development mode though, it should never be used in production.

_posts/2019-06-05-quarkus-and-web-ui-development-mode.adoc
kabir: No, this is only needed when running a packaged angular application, this is unrelated to the proxy.

Say you have an angular route called /other like in the example. If you load this from angular CLI it works ok (as angular knows what you are doing). If you are not familiar with routes in modern web applications, https://angular.io/guide/router might help.

However, when deployed in Quarkus (or any other app server for that matter) it will look for a resource called /other. There is no file called that, no REST endpoint and no servlet called that either. So the server will give a 404 error. 

All the server knows about from the angular side is that there is an index.html which pulls in the transpiled front-end resources. 

So in the case of something not found, we do a **forward** to the default location of the application, which loads that index.html and retains the path and query parameters to be parsed by Angular on the front-end. 

This allows us to bookmark pages. Without this you can only go to localhost:8080 (which loads up the index.html), but not directly to something handled by Angular like localhost:8080/other. This is the only time the filter is actually needed.

kabir: I can try to explain a bit better in that section I guess

kabir:
@emmanuelbernard I have updated it. Unless my latest changes introduced something weird you pick up on, it is good to go from my side

rsvoboda:
Merged, there we no additional comments to this PR nor the original #210

emmanuelbernard:
Thanks @kabir I tweeted it
https://twitter.com/QuarkusIO/status/1136971830873481216

PR #231, opened 2019-06-18
Generate Guides page from a YAML file

gsmet:
And a minor fix to a blog post.


gsmet:
@cescoffier could you take a look? It was a fun ride...

Issue #244, opened 2019-06-27
Warning About OpenJDK 12 for Step 2 of QUARKUS - GET STARTED


danielhelfand:
Quarkus [does not support OpenJDK 12](https://github.com/quarkusio/quarkus/issues/1534) at this time, but it's an option after clicking on the link for `You need JDK 8+` as shown [here](https://adoptopenjdk.net/).

It would be helpful to provide some type of warning to only download OpenJDK 8 or 11 at this time to help others avoid the issue above.

emmanuelbernard:
Fixed it in an overall.

Issue #240, opened 2019-06-27
Get Started Step 1 - You need an IDE Should List Java IDEs Instead of Going to IDE Wikipedia Page


danielhelfand:
As part of the getting started information, it would be more helpful to list specific Java IDEs that might be helpful instead of linking to a general Wikipedia page on IDEs for all languages.

The source in question is [here](https://github.com/quarkusio/quarkusio.github.io/blob/15009263af366d392044032dcb1c1a5cdb26685d/_includes/get-started-band.html#L5). Simply changing the link to https://en.wikipedia.org/wiki/Comparison_of_integrated_development_environments#Java would be more helpful, but it might be better to be a little more particular about recommendations or find a better list than what's on Wikipedia.

danielhelfand:
For the time being, I have updated the link to https://en.wikipedia.org/wiki/Comparison_of_integrated_development_environments#Java, but I'm wondering if there are better suggestions than this list or if general there are better ideas for what to link this to.

emmanuelbernard:
Thanks @danielhelfand I was in this area and did an overhaul. Let me know.

danielhelfand:
@emmanuelbernard Appreciate it for this and #244. It's a lot more understandable with better information but still has brevity to it. Nice work!

PR #245, opened 2019-06-28
Issues 218 240 244 - Show Quarkus version in Getting Started page

### LINKED ISSUE #218: Offer on the home page or somewhere else the latets release version and link to the shownotes

emmanuelbernard:
Clarify JDK version supported
Clarify GraalVM version
Clarify GraalVM is optional
Clarify IDE choices

Fixes #218 #240 #244

### DIFF
--- a/_data/versions.yaml
+++ b/_data/versions.yaml
@@ -1,0 +1,7 @@
+quarkus:
+  version: 0.18.0
+  announce: /blog/quarkus-0-18-released/
+graalvm: 19.0.2+
+jdk: "8 to 11"
+maven: 3.5.3+
+
--- a/_includes/get-started-band.html
+++ b/_includes/get-started-band.html
@@ -11,10 +11,10 @@
     <div class="number"><div>2</div></div>
     <div class="text">
       <div class="text">
-        You need <a href="https://adoptopenjdk.net" target="_blank">JDK 8 to 11</a>
+        You need <a href="https://adoptopenjdk.net" target="_blank">JDK {{ site.data.versions.jdk }}</a>
       </div>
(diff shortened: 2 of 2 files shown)


_data/versions.yaml
```diff
@@ -0,0 +1,7 @@
+quarkus:
+  version: 0.18.0
+  announce: /blog/quarkus-0-18-released/
+graalvm: 19.0.2+
+jdk: "8 to 11"
+maven: 3.5.3+
```
rsvoboda: I thought we require 3.5.4+ due to changes @aloubyansky worked on.

emmanuelbernard:
If that's the case then that's a fail from code change to doc propagation
because the website says 3.5.3

rsvoboda:
Our .adoc files in main Quarkus say `3.5.3+` so the docs propagation process is not broken, probably our .adoc files need update. Talking about it on Zulip now.

rsvoboda:
mvn 3.5.3 worked well on both quarkus and the quickstarts

My motivation for this fuzz was the fact that https://github.com/quarkusio/quarkus/blob/master/bom/pom.xml#L60 says `<maven-core.version>3.5.4</maven-core.version>`

_data/versions.yaml
```diff
@@ -0,0 +1,7 @@
+quarkus:
+  version: 0.18.0
+  announce: /blog/quarkus-0-18-released/
+graalvm: 19.0.2+
+jdk: "8 to 11"
```
gsmet: It's a detail but I would say `8 or 11`.

This makes it clear that we only test and support the LTS versions.

Issue #213, opened 2019-05-30
Improve description in RSS feed


emmanuelbernard:
In the RSS description field, you pass HTML. I wonder if it would be better to remove HTML tags altogether and just get raw text.
Especially since we cut the content at 400 characters, we will get non closed tags and all.

PR #287, opened 2019-09-05
Fix invalid XML in feed

pascalw:
Hi there! I wanted to follow your RSS feed but it could not be parsed by my RSS reader due to invalid XML (see [feed validator](https://validator.w3.org/feed/check.cgi?url=https%3A%2F%2Fquarkus.io%2F%2Ffeed.xml)).

`post.content` contains HTML. Truncating it arbitrarily at 400 chars
could result in raw `&` characters for example to end up in the feed.

This commit simply strips HTML and then truncates as suggested in #213.

### DIFF
--- a/feed.xml
+++ b/feed.xml
@@ -17,7 +17,7 @@ layout: null
                 {{ post.url | prepend: site.url }}
             </link>
             <description>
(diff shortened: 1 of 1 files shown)


gsmet:
Makes sense, thanks!

PR #294, opened 2019-09-23
Replace "application/json" with MediaType.APPLICATION_JSON

ge0ffrey:


gsmet:
This should be done in the main repo as the files are copied from there.

PR #297, opened 2019-09-25
Better use of mobile space

### LINKED ISSUE #295: mobile layout for site and blog don't look good
The main content doesn't use the full with of the page
Screenshot from 2019-09-25 10-36-15
Embedded video takes up more size then available on mobile:
Screenshot from 2019-09-25 10-37-18
There is a lot of space used for the icon not a lot for the text for admonitions on mobile:
Screenshot from 2019-09-25 10-36-49

edewit:
fixes #295

### DIFF
--- a/_sass/core/global.scss
+++ b/_sass/core/global.scss
@@ -19,6 +19,16 @@ body > div.content {
   padding: 0 13rem;
   @media screen and (max-width: 1170px) { padding: 0 6rem; }
   @media screen and (max-width: 480px) { padding: 0 2rem; }
+  @media screen and (max-width: 768px) {
+    padding: 0 1.2rem;
+    .width-10-12 {
+      grid-column: span 12;
+}
+    .admonitionblock > table td.icon {
+      width: initial;
+      font-size: 75%;
+  }
+}
 }
 
 .hide {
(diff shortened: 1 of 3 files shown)


edewit:
Result:
Screenshot from 2019-09-25 10-52-35

_sass/core/global.scss
```diff
@@ -19,6 +19,16 @@ body > div.content {
   padding: 0 13rem;
   @media screen and (max-width: 1170px) { padding: 0 6rem; }
   @media screen and (max-width: 480px) { padding: 0 2rem; }
+  @media screen and (max-width: 768px) {
+    padding: 0 1.2rem;
+    .width-10-12 {
+      grid-column: span 12;
+    }
```
gsmet: I wonder if we are making good use of the grid system because this sort of things should be automatic.

edewit: It looks fine on desktop using only 10 columns of the 12, but on mobile you want to use all of them

gsmet: So the right fix would be to add `width-12-12-m` on the div in addition to `width-10-12`. That will solve the issue. Let's avoid fiddling with the grid system :).

BTW, we should also probably adjust `.navigation-wrapper` for mobile and tablet. Would you be interested in having a look?

edewit: right there seems to be some big padding below it, 63e01117ea8aaa2e2f307462d5051f760142c76e should fix that

gsmet:
Merged, thanks a lot!

PR #306, opened 2019-10-02
Add missing path for UserResource

AlexanderHofmeister:
I went through the keycloak setup and noticed the missing path of the UserResource.


gsmet:
This type of changes needs to be pushed to the main repo but as we also renamed the keycloak guide, I will apply it there.

Thanks for taking the time to report it!

PR #310, opened 2019-10-09
Get ready for config tables

FroMage:
Added links to config js/css and fixed some css

### DIFF
--- a/_layouts/base.html
+++ b/_layouts/base.html
@@ -13,6 +13,7 @@
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <meta name="description" content="{% if page.description %}{{ page.description }}{% else %}{{ site.description }}{% endif %}">
   <link rel="shortcut icon" type="image/png" href="{{ "/favicon.ico" | prepend: site.baseurl }}" >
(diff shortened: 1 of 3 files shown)


gsmet:
- [ ] `g` on first line cut
- [ ] Font of the duration admonition too big
- [ ] Vertical line on the admonition lost
- [ ] name of property and first line of description are not aligned
- [ ] if you go on http://localhost:4000/guides/hibernate-search-guide#duration-note-anchor then search for `analysis`, table is not 100% in width anymore
- [ ] we need to fix the titles on the guides already including the doc otherwise we have two (and we need to tweak the title level offset too): `Hibernate Search Configuration Reference` and `Configuration options` here: http://localhost:4000/guides/hibernate-search-guide#duration-note-anchor

FroMage:
>  if you go on http://localhost:4000/guides/hibernate-search-guide#duration-note-anchor then search for analysis, table is not 100% in width anymore

Yes, and if I fix this properly by setting the table width to be `100%` then the columns sizes are wrong, even with them being specified in `colgroup` AND on each `td`. I considered the fix worse than this issue. But if you know how to fix it let me know.

FroMage:
All done, but you need to repull https://github.com/quarkusio/quarkus/pull/4381 as well as this PR.

PR #313, opened 2019-10-13
Delete keycloak guide

madmas:
as of the replacement of keycloak with openid-connect extension (see https://quarkus.io/blog/quarkus-0-24.0-released/#keycloak-extension-replaced-by-an-oidc-extension-openid-connect) and the new guide already available at https://quarkus.io/guides/oidc-guide, this guide should be deleted as it's obsolete and due to the non.working commends and references misleading.


cescoffier:
Thanks!

PR #318, opened 2019-10-21
Hide TOC when content is empty

ia3andy:
A few tweak as part of restyling the all config with @FroMage.


emmanuelbernard:
@ia3andy is that still relevant, I know the all config has evolved a lot recently

ia3andy:
@emmanuelbernard yes, it's a trick so the TOC is not displayed when empty

PR #327, opened 2019-10-31
Guide cleanup

gsmet:
This should not be merged before the new guides are synced up.


_config.yml
```diff
@@ -18,7 +18,7 @@ email:
 description: >- # this means to ignore newlines until "baseurl:"
   Quarkus: Supersonic Subatomic Java
 baseurl: "" # the subpath of your site, e.g. /blog
-url: "https://quarkus.io/" # the base hostname & protocol for your site, e.g. http://example.com
+url: "https://quarkus.io" # the base hostname & protocol for your site, e.g. http://example.com
```
gsmet: This was an error, it shouldn't contain the trailing `/`.

gsmet:
@cescoffier could you have a look: it's just something to get proper redirects from the old guides to the new ones. I generated the files from the git diff so it should be safe enough.

PR #328, opened 2019-11-01
Fix: rest-client-multipart guide

hectorvent:
Fix: wrong values assignments in rest client guide.


gsmet:
Good catch! The file is maintained in the main repo here: https://github.com/quarkusio/quarkus/blob/master/docs/src/main/asciidoc/rest-client-multipart.adoc (I recently renamed it) and then copied to the website when we release.

Could you create a PR against the main repo?

Thanks!

hectorvent:
Yes, I will.

gsmet:
Thanks!

PR #353, opened 2019-11-06
Canonical headers

### LINKED ISSUE #352: Content duplication - SEO penalties
Most pages can be linked either with or withouth the `.html` postfix.

For example, both these links show the same content:

- https://quarkus.io/guides/cdi-reference.html
- https://quarkus.io/guides/cdi-reference

This gets classified as content duplication and ambiguous canonical source, which will get us severe penalties in term of SEO scores.

I see two possible fixes:

1. Add a "canonical" header to the content, so to at leat flag which page is meant to be the canonical one. It's acceptable to refer to "self" as well so this could easily be included in the page.
2. Remove the pages having the `.html` postfix - and possibly redirect?

Sanne:
Also needed to fix the guides not having a permalink, or the preference was being given to the links with .html postfix.

Fixes #352

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -68,6 +68,7 @@ defaults:
       type: "guides" # previously `post` in Jekyll 2.2.
     values:
       layout: "guides"
+      permalink: /guides/:name
   -
     scope:
       type: "posts"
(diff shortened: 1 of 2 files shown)


gsmet:
Tested and looks good. I created a PR for the website to move all the cross-guides links to the canonical form: https://github.com/quarkusio/quarkus/pull/5258 .

Sanne:
thanks!

PR #355, opened 2019-11-06
Make right hero centered

FroMage:


gsmet:
@jkbrock @adelasofia can you check if it's OK with you?

jkbrock:
@gsmet it looks good to me. Go for it.

Issue #360, opened 2019-11-07
Old Quarkus URLs are indexed by Google


gwenneg:
I just searched `quarkus vertx guide` with Google and the first link from the result page was https://quarkus.io/guides/using-vertx.html which no longer exists since the guide file was renamed recently. Instead of accessing the guide, I ended up seeing a Quarkus 404 page.

This link will be removed from the Google index eventually, but this process can be sped up by submitting a removal request using https://www.google.com/webmasters/tools/removals

I just did it for the Vert.x guide, but this should also be done for all Quarkus guides that were renamed. Actually, it might even be a good idea to check all Quarkus URLs indexed by Google with a few of the most common Quarkus keywords combinations.

gsmet:
So the issue is a bit more complicated than that: there are redirections
but apparently they are not working when you use the .html form (that
shouldn't be used but we had links to .html version before).

https://quarkus.io/guides/using-vertx redirects properly to
https://quarkus.io/guides/vertx but not the .html.

Yoann is trying to find a work around.

gsmet:
By the way, you have a list of all the redirected guides here:
https://github.com/quarkusio/quarkusio.github.io/tree/develop/_redirects/guides
.

So you can easily get the list of pages to remove from Google.

gwenneg:
I didn't know about the redirection, that makes the problem completely different indeed.

Thanks for the info!

gwenneg:
Fixed with #363

PR #363, opened 2019-11-08
Fix redirects from *.html pages

yrodiere:
Apparently some .html URLs were indexed by Google, and we're getting 404 status pages when clicking on the Google search results. This fixes the problem.


gwenneg:
Fixes #360

PR #368, opened 2019-11-13
Blog on MP 3.2 compatibility with Quarkus

kenfinnigan:


kenfinnigan:
After chatting with John, going to ice this for now

PR #369, opened 2019-11-13
Correct GraalVM related items in the glossary 

shelajev:
Updated Glossary with more correct explanations for the GraalVM technology terms


cescoffier:
Thanks.

gsmet:
FYI, this is just an internal glossary. It's not linked anywhere in the website.

PR #371, opened 2019-11-14
Fixes https://github.com/quarkusio/quarkus/issues/1942

FroMage:
Link client/server rest guides
Make sure searching for json on the guides page will return client and server


gsmet:
@FroMage did you create a PR for the guides on the main repo? Because they will be overwritten by the next release otherwise. I don't think I have seen it and it would need to be backported too.

FroMage:
Totally: https://github.com/quarkusio/quarkus/pull/5465

PR #373, opened 2019-11-18
2. newsletter with reimplemented pub system

maxandersen:
Why:

 * time for a new newsletter
 * scholar seem to fail on queries

This change addreses the need by:

 * added new stories and a blog entry
 * removed scholar and replaced with simplest
   change to mimic scholar but simpler
 * migrated bibtex to a yaml file.


maxandersen:
should be straightforward and it actually seems about 1s faster so that’s good ;)

maxandersen:
I'll add a few more articles I just got - but do please review/verify the new generation works.

maxandersen:
merged to get airtime before CR2/Final :)

PR #405, opened 2019-12-19
Added blog articles

fmarchioni:
Added a couple of articles to the list of publications


maxandersen:
missed this over xmas break - mergin in now and will be in news #4

PR #497, opened 2020-04-02
Adding log level

agoncal:
I think it's important to mention all the log levels. There are so many loggers that we might get confused (eg. TRACE vs FINEST, etc.)


gsmet:
This needs to be done in the main repo or it will be lost at the next release.

See the header of the file you modified for more info.

agoncal:
Got it. https://github.com/quarkusio/quarkus/pull/8360

PR #595, opened 2020-05-21
Duplicated Kubernetes client guide

torrespro:


emmanuelbernard:
Thanks @torrespro, applied

PR #635, opened 2020-06-15
Fix invalid feed.xml generation

djnalluri:
The inclusion of '&' in the title is generating an invalid feed.xml as it appears to not be correctly escaped.

### DIFF
--- a/_posts/2020-05-20-insights-4-5.adoc
+++ b/_posts/2020-05-20-insights-4-5.adoc
@@ -1,6 +1,6 @@
 ---
 layout: post
-title: 'Quarkus Insights #4 & #5: JHipster and Quarkify'
+title: 'Quarkus Insights #4 and #5: JHipster and Quarkify'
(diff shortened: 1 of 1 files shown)


emmanuelbernard:
FYI @maxandersen don't use & in titles.

emmanuelbernard:
thanks @djnalluri , applied.

maxandersen:
We are generating feeds without urlrncoding the generated content ? That's not good. I'll need to stop using # too then :)

emmanuelbernard:
Not for the title. It's a common issue around static website generators

maxandersen:
> Not for the title. It's a common issue around static website generators

[not true](https://github.com/quarkusio/quarkusio.github.io/pull/636) - it s a common issue for not proper made templates when original templates were fixed [7+ years ago](https://github.com/jekyll/jekyll/pull/1501) :)

PR #636, opened 2020-06-18
xml escape feed more

maxandersen:
since I want to be able to use proper ligature.


emmanuelbernard:
I applied but we should apply a RSS lint to make sure it's fine

emmanuelbernard:
I checked using https://validator.w3.org/feed/
We have a small unrelated warning

PR #641, opened 2020-06-26
add a border to all code blocks, even those without hightlightjs

ahus1:
The code blocks that don't receive highlightjs miss a border and don't use standard white font color like for example shell script.

Example: https://quarkus.io/guides/vault-datasource

AsciiDoc (note the missing `[source]`): 
```
----
mvn quarkus:add-extension -Dextensions="io.quarkus:quarkus-hibernate-orm,io.quarkus:quarkus-jdbc-postgresql"
----
```

Rendered in browser before change: 

image

Rendered in browser after change:

image

### DIFF
--- a/_sass/quarkus.scss
+++ b/_sass/quarkus.scss
@@ -301,11 +301,12 @@ pre {
  padding: 0;
 }
 
-pre.highlight {
+pre.highlight, .literalblock pre, .listingblock > .content > pre {
  border: 1px solid $quarkus-blue;
  padding: 1rem;
  line-height: 1.2em;
  overflow-x: auto;
+ color: #EFEFEF;
 }
 
 pre > code,


emmanuelbernard:
Isn't the better fix to add the `[source]` element to the vault data source code?

emmanuelbernard:
I'm saying that because I'm wondering where else 

- pre.highlight is used
- content > pre is used

ahus1:
@emmanuelbernard - I've been thinking about adding [source] to all source code/command line example snippets, but decided against it: 

* in the usual AsciiDoc style sheet adding this type doesn't change the output a writer will see in their preview
* writers will be used to add it only when they add also a source code language
* there is no check enforced (like for example a linter) that will prevent this to be missed in future changes or other files

Therefore I went for the CSS change. 

After reading your comment I visited the standard [Asciidoctor stylesheet](https://github.com/asciidoctor/asciidoctor/blob/master/data/stylesheets/asciidoctor-default.css#L208) to find out if I could make the CSS more specific. 

This is the snippet that adds the border for all listing in the default Asciidoctor stylesheet:

```
.literalblock pre,.listingblock>.content>pre
{-webkit-border-radius:4px;border-radius:4px;word-wrap:break-word;overflow-x:auto;padding:1em;font-size:.8125em}
```

I updated the CSS in this PR to make it as specific as the Asciidoctor default stylesheet. I manually tested it and it creates the same screenshot as above.

Therefore I suggest this PR for a CSS change.

gsmet:
FWIW, the type is not needed to have a proper style. But you need the `[source]` block.

But if Asciidoc works correctly, we should probably fix the CSS as that's not something you can see when checking the output.

emmanuelbernard:
@ahus1 understood.
It hink your change would scare me a bit less if you were also keeping the old `pre.highlight` rule vs replacing it with the new set you propose. Is there a reason you decided to drop `pre.highlight` from the rule set?

ahus1:
@emmanuelbernard - I tried to keep it minimal: my tests showed that for this and some manually selected other documentation examples it wasn't necessary.

I can add `pre.highlight` to the CSS to be safe.

emmanuelbernard:
@ahus1 yes let's add it back and ship this one. Note that Adela did some CSS work, let's rebase this PR before going

ahus1:
@emmanuelbernard - ok, `pre.highlight` re-added. Ready for another review.

emmanuelbernard:
Thanks @ahus1 and sorry for the review turnaround

PR #643, opened 2020-06-30
Refactors the header navigation

adelasofia:
Completely reworked the header navigation for a better experience on all screen sizes. L&F approved by James.


emmanuelbernard:
Clever. Thanks Adela

PR #672, opened 2020-07-29
Update march of IDEs features set of IntelliJ IDEA Ultimate

jreznot:
IntelliJ IDEA 2020.2 supports new project wizard and configuration files: .properties and YAML

https://www.jetbrains.com/idea/whatsnew/#frameworks--technologies

### DIFF
--- a/_posts/2020-01-21-march-of-ides.adoc
+++ b/_posts/2020-01-21-march-of-ides.adoc
@@ -92,14 +92,14 @@ https://download.jboss.org/jbosstools/intellij/snapshots/intellij-quarkus/[Devel
 |icon:check[]
 |icon:check[]
 |https://issues.jboss.org/browse/JBIDE-26950[icon:times[]]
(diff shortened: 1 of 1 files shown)


emmanuelbernard:
can I let you lead that @maxandersen BTW I don't think this blog entry has a ide tag.
Also, might be good to do a round of IDE invites on Insights :)

emmanuelbernard:
thanks for the update @jreznot 
This is an old blog, so I was wondering how we can give the improvements more visibility to the community (on top of the fixes you gave)

jreznot:
> thanks for the update @jreznot
> This is an old blog, so I was wondering how we can give the improvements more visibility to the community (on top of the fixes you gave)

Yes, that could be a new blog ) But for now I just want to set correct up-to-date marks in case users find the blog post.

gsmet:
Merged, thanks!

PR #675, opened 2020-08-03
Updates guides page

adelasofia:
In response to: https://issues.redhat.com/browse/DESIGN-1947

Updates https://quarkus.io/guides/ page with new styles and search function.

### DIFF
--- a/_includes/back-to-top.html
+++ b/_includes/back-to-top.html
@@ -1,0 +1,1 @@
+<a id="backToTop"><i class="fas fa-chevron-up"></i></a>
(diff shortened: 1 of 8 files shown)


gsmet:
Thanks for that! It's so much better than what we have that I will merge it right away.

There are a couple of things we could maybe improve:
* could you make the scroll when clicking a category smoother? You have done it for the back to top but not when clicking a category. That usually helps understanding what's happening.
* there are a few descriptions we need to adjust as they are far too long, I will do that.

PR #784, opened 2020-11-05
Try to switch to another type of node

gsmet:
Lately, we had huge delay for build execution, let's see if using a
more recent type of node helps.


gsmet:
Doesn't change anything :/

PR #751, opened 2020-09-29
First version of the daily push of the web site

cescoffier:

PR #790, opened 2020-11-11
Replaces Travis with GitHub Actions

gastaldi:
This removes the Travis descriptor and builds using GitHub actions. It also closes #751 by building daily.

This is the recommended approach as in https://jekyllrb.com/docs/continuous-integration/github-actions/

Travis has enforced some limits recently (see https://www.theregister.com/2020/11/02/travis_ci_pricng/).


.github/workflows/github-pages.yml
```diff
@@ -0,0 +1,19 @@
...
+  push:
+    branches:
+      - develop
+
+jobs:
+  github-pages:
+    runs-on: ubuntu-16.04
+    steps:
+      - uses: actions/checkout@v2
+      - uses: helaili/jekyll-action@2.0.4
```
gsmet, line 16: Isn't there an official GH one?

.github/workflows/github-pages.yml
```diff
@@ -0,0 +1,19 @@
+name: Build and deploy to GitHub Pages
+
+on:
+  schedule:
+    - cron: '0 0 * * *'
+
+  push:
+    branches:
+      - develop
```
gsmet, line 9: I don't see PRs being tested? Travis was doing that IIRC.

.github/workflows/github-pages.yml
gastaldi: No, if you look at the [official docs](https://jekyllrb.com/docs/continuous-integration/github-actions/) you'll see that's the recommended action

.github/workflows/github-pages.yml
gastaldi: Ah yes, I'll have a look

gastaldi: Preferably it would be nice if PRs could deploy to surge.sh to preview how the website would look like

gastaldi:
Moving to a branch in the same repository

PR #798, opened 2020-11-16
Add some keywords for the CDI-related guides

gsmet:
/cc @mkouba @gastaldi @geoand I added this feature recently. I added some keywords for the SSO-related guides and in this PR I add some for CDI (I was looking for things in the CDI doc).

If you think of others, be my guest. They are not visible but are included in the search engine.

Obviously, we don't want too many either as we don't want all the guides to pop up when looking for something but some others might be useful.

It's not pressing but I thought it could be useful to make you aware of it.

### DIFF
--- a/_data/guides.yaml
+++ b/_data/guides.yaml
@@ -32,10 +32,11 @@ categories:
       - title: Introduction to CDI
         url: /guides/cdi
         description: Quarkus DI solution is based on the [Contexts and Dependency Injection for Java 2.0](https://docs.jboss.org/cdi/spec/2.0/cdi-spec) specification. This guide explains the basics of CDI.
-        keywords: qualifier event interceptor observer
+        keywords: qualifier event interceptor observer arc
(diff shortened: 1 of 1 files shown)


geoand:
Thanks for the heads up!

Really useful indeed!

mkouba (approved):
LGTM.

OT: I'd like to show the `cdi` and `cdi-reference` guides when `arc` is used in the search box. What should I do? Do I need to modify the description, or?

gastaldi:
> OT: I'd like to show the `cdi` and `cdi-reference` guides when `arc` is used in the search box. What should I do? Do I need to modify the description, or?

I think you just need to add the `keywords: arc` to these guides

gsmet:
Exactly, that's the whole point of this. We can add keywords without having to make the description hard to read.

Thanks @gastaldi for adding `arc`, let's merge this!

PR #913, opened 2021-03-05
Fix typo in the article 'Quarkus 1.12.1.Final released - Bugfixes'

ynojima:
In the article 'Quarkus 1.12.1.Final released - Bugfixes', 
It is stated that '1.12.1.Final is a safe upgrade for everyone using Quarkus 1.11.' but it seems it should be '1.12.1.Final is a safe upgrade for everyone using Quarkus 1.12.'


gsmet (approved):
Good catch, thanks!

PR #936, opened 2021-03-16
Adding a second GA tag to take advantage of GA4's capabilities

insectengine:

### DIFF
--- a/_data/insights-videos.yaml
+++ b/_data/insights-videos.yaml
@@ -1,0 +1,259 @@
+futureheadline: "Upcoming Podcasts"
+
+futurevideos:
+- title: "EPISODE #46: Quarkus and AWS Lambda Containers"
+  link: 
+  date: "Aprl 26, 2021"
+  authors: "Aaron Walker"
+
+- title: "EPISODE #45: Writing github bots with Quarkus"
+  link: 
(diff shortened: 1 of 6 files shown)


maxandersen:
Why do we need the second one ? Shouldnt one be sufficient ?

Any of these slightly slows down the page so let's be careful.

insectengine:
@maxandersen  -  The original GA number is owned by GoldenGate and it's controlled by RHD. We've gotten accidentally locked out in the past. We can't transfer that number to a new account and maintain the historical data (as per Axel). We are adding this new number to take advantage of GA's new reports in Google Analytics 4.  

https://support.google.com/analytics/answer/10089681?hl=en

gastaldi:
Looks good: https://quarkusio-quarkusio-github-io-preview-pr-936.surge.sh/insights/

maxandersen:
@insectengine seems two prs in one. its just the goan.js you want applied, right James?

insectengine:
> @insectengine seems two prs in one. its just the goan.js you want applied, right James?

Correct... we cherry picked the change to the QI changes to another commit.

maxandersen:
i've merged just the goan.js file and checked it with tagmanager extension that both tags are detected now.
 
where can I see the the new info ?

Issue #941, opened 2021-03-14
The All Config page is incredibly slow


richiethom:
## Describe the bug
https://quarkus.io/guides/all-config is incredibly slow to load.

### Expected behavior
A page that loads quickly that I can then filter/search.

### Actual behavior
A  page that loads very slowly, sometimes not at all. If it does load, then the results are still search/filterable.

mkouba:
It seems that a great deal of time is spent in the `config.js`, namely when iterating over the nodes with CSS class `description` (more than 2500 occurrences): https://github.com/quarkusio/quarkusio.github.io/blob/develop/_guides/javascript/config.js#L24-L26

mkouba:
I think that we can close this one. The page loads reasonably fast now.

PR #942, opened 2021-03-19
Attempt to reduce the number of style recalc

mkouba:
- related to #941

This could help a little. Unfortunately, I can't test the whole site locally (no functional ruby env here).

### DIFF
--- a/_guides/javascript/config.js
+++ b/_guides/javascript/config.js
@@ -18,13 +18,23 @@ if(tables){
           caption.children.item(0).appendChild(input);
           input.addEventListener("keyup", initiateSearch);
           input.addEventListener("input", initiateSearch);
-          inputs[input.id] = {"table": table};
           var descriptions = table.querySelectorAll(".description");
           if(descriptions){
+            var heights = new Array(descriptions.length);
+            var h = 0;
             for (description of descriptions){
-              makeCollapsible(input, description);
+              heights[h++] = description.offsetHeight;
             }
+            var shadowTable = table.cloneNode(true);
+            var shadowDescriptions = shadowTable.querySelectorAll(".description");
+            h = 0;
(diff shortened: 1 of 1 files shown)


mkouba:
We could also try to skip this part completely: https://github.com/quarkusio/quarkusio.github.io/blob/develop/_guides/javascript/config.js#L257-L260

FroMage:
Actually, I believe we need to work on the shadow table for this to work. Lemme give it a try. But damn I had garbage collected the memory of this :(

FroMage:
I've done this with a shadow DOM, at the cost of some false positives, but the time required for setting it up is now down to 250ms on my machine. We should have done this long ago, but initially there was no need since we had much fewer entries.

The problem is that our script doesn't get called until quite a long time, because _something_ is taking forever and delays the page load. I'm not sure what, though.

FroMage:
Note that this actually needs fixing in the quarkus repo, not just here.

FroMage:
@mkouba @gsmet we should merge this somehow, no?

mkouba:
> @mkouba @gsmet we should merge this somehow, no?

+1 I've tried the the [preview page](https://quarkusio-quarkusio-github-io-preview-pr-942.surge.sh/guides/all-config) and it seems to load faster...

gsmet:
@FroMage I can merge it for sure. I'll let you fix it in the Quarkus repo as you said it was needed?

FroMage:
> @FroMage I can merge it for sure. I'll let you fix it in the Quarkus repo as you said it was needed?

well, yeah, that's the source, when you make a release, no?

gsmet:
Yes if this file is coming from the main repo, it needs an update before the next release (probably next Wednesday).

FroMage:
Done: https://github.com/quarkusio/quarkus/pull/16436

PR #971, opened 2021-05-03
Additional Config guides.

radcortez:
Requires https://github.com/quarkusio/quarkus/pull/16496

### DIFF
--- a/_data/guides-main.yaml
+++ b/_data/guides-main.yaml
@@ -28,6 +28,15 @@ categories:
       - title: Configuration Reference
         url: /guides/config-reference
         description: Learn more about how to configure your Quarkus applications.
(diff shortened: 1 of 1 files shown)


gsmet:
@radcortez any chance you could update https://github.com/quarkusio/quarkusio.github.io/blob/develop/_data/guides-main.yaml instead? This way, I will be able to merge it right away and they will be accessible if you switch to the main branch.

radcortez:
> @radcortez any chance you could update https://github.com/quarkusio/quarkusio.github.io/blob/develop/_data/guides-main.yaml instead? This way, I will be able to merge it right away and they will be accessible if you switch to the main branch.

Sure. Sorry, I thought it was the other file.

PR #983, opened 2021-05-18
Fix relative path to properly display version select list

soul2zimate:
some guides versions list is broken.

e.g. https://quarkus.io/guides/building-my-first-extension versions list is broken.


gsmet (approved):
Good catch, thanks for the fix!

PR #1015, opened 2021-06-18
@PathParam needs default value: javax.ws.rs.PathParam missing element…

closet6:
@PathParam needs default value: javax.ws.rs.PathParam missing element…


closet6:
further testing suggests a better way

PR #1042, opened 2021-07-01
Correct markup error

ynojima:
This mark-up error breaks ja.quarkus.io build and need to be fixed.
https://github.com/quarkusio/ja.quarkus.io/actions/workflows/sync.yml


gsmet (approved):
Good catch!

PR #1043, opened 2021-07-01
Replace old attributes

ynojima:
`allowfullscreen` attribute also breaks [ja.quarkus.io](https://github.com/quarkusio/ja.quarkus.io) build because it cannot be parsed by po4a parser. 
It would be appreciated if it is replaced with `allow="fullscreen"`

### DIFF
--- a/_layouts/quarkus2.html
+++ b/_layouts/quarkus2.html
@@ -84,13 +84,13 @@ layout: vision
     <h3>Quarkus Insights Episode #55: The Quarkus 2.0 Launch</h3>
   </div>
   <div class="width-12-12 video-container">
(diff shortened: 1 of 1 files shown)


gastaldi (changes requested):
Unfortunately that doesn't work. Maybe it needs to be appended to the existing `allow` attribute?

gastaldi:
Or use `allowfullscreen="allowfullscreen"` instead

ynojima:
Oh, I overlooked existing one, sorry.

gastaldi (approved):
I can confirm that works now. Thanks!

PR #1050, opened 2021-07-06
Add newsletter page, first newsletter (Issue #1 - July 2021)

insectengine:


maxandersen:
1) newsletter content does not seem to match what shown on gchat issues like attributions for articles are missing,font in first paragraph missing and  I now noticed the blurp from jason missing?
2) can we rename the newsletter files to be like folder 10 with index.html so instead of quarkus.io/newsletter/newsletter_10_0721.html urls we get quarkus.io/newsletter/10  

i tried finding a good way of havign automatic blog entry per newsletter but realized we might want to publish newsletter seperately from Quarkus newsletter so having a blog entry with link might be just as fine.

PR #1116, opened 2021-08-24
fix: add author to feeed

maxandersen:
Wanted to be able to fetch from /feed.xml posts and have author tag to filter/sort by.

this just adds standard author tags to the feed.


maxandersen:
also - no idea why we have both jekyll-feed plugin enabled and then have feed.xml that overwrites it but thats another pr in future ;)

PR #1118, opened 2021-08-25
Updated the Quarkus repo URL (master -> main)

paulrobinson:
It's mentioned in many places to use the https://github.com/quarkusio/quarkus/tree/master/docs/src/master/asciidoc URL for contributing guides. That URL has now changed to https://github.com/quarkusio/quarkus/tree/master/docs/src/main/asciidoc.

This PR updates that URL in the docs, and in the headers of all the versioned guides. 

See https://source.redhat.com/aboutredhat/weareredhat/inclusion/diversity__inclusion_wiki/conscious_language_project_faq for the reasons why we are making these changes and for guidance.


gsmet (approved):
Yeah, it's mostly the doc coming from the old branches but good to fix anyway. Thanks!

paulrobinson:
Thanks @gsmet. I figured the old docs wouldn't be updated again by the script, so was worth changing.

PR #1119, opened 2021-08-25
Updated the quarkus-tree-url variables (master -> main)

paulrobinson:
Updating the `quarkus-tree-url` variable to point to the updated Quarkus repo URL, that has renamed the `master` branch to `main`.

I've not checked every usage of the `quarkus-tree-url` variable, but I don't see any situations in which the rename would break. 

See https://source.redhat.com/aboutredhat/weareredhat/inclusion/diversity__inclusion_wiki/conscious_language_project_faq for the reasons why we are making these changes and for guidance.


gsmet (approved):
Good catch!

PR #1125, opened 2021-08-31
More 'master' => 'main' fixes

paulrobinson:
Here's some more 'master' to 'main' fixes. I think we're now getting close to having all the ones that need changing done.

### DIFF
--- a/_guides/cassandra.adoc
+++ b/_guides/cassandra.adoc
@@ -52,7 +52,7 @@ We recommend that you follow the instructions in the next sections and create th
 by step. However, you can go right to the completed example.
 
 The solution is located in the
-link:https://github.com/datastax/cassandra-quarkus/tree/master/quickstart[quickstart directory] of
+link:https://github.com/datastax/cassandra-quarkus/tree/main/quickstart[quickstart directory] of
 the Cassandra Quarkus extension GitHub repository.
 
 == Creating a Blank Maven Project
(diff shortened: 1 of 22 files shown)


_guides/cassandra.adoc
```diff
@@ -52,7 +52,7 @@ We recommend that you follow the instructions in the next sections and create th
 by step. However, you can go right to the completed example.
 
 The solution is located in the
-link:https://github.com/datastax/cassandra-quarkus/tree/master/quickstart[quickstart directory] of
```
gastaldi, line 55: Actually this is already updated in https://github.com/quarkusio/quarkus/blob/main/docs/src/main/asciidoc/cassandra.adoc.

I am not sure when it is synch'ed?

paulrobinson: Ah yes, I see that. 

Is this a sign that the sync is not correctly happening? I would have expected it to happen after every release of Quarkus. So looking at https://github.com/quarkusio/quarkus/tags, I'd have expected the sync to happen with the 2.2.3.Final release 10 days ago.

gsmet: Depends if it has been backported or not.

paulrobinson: Ah yes, of course.

Should we just leave this change in, and let it be over-written when 2.3.0.Final is released? Or would you prefer I removed this and look for any similar cases?

gastaldi: I'd prefer to have this change in a PR in the quarkusio repo with a backport label, but that's me :)

gastaldi: But if 2.3.0 is coming soon maybe it's not necessary

paulrobinson: 2.3.0.Final is estimated for Sept 29th, so it's soon enough IMO for this type of change. See: https://github.com/quarkusio/quarkus/wiki/Release-Planning

PR #1206, opened 2021-11-12
update examples of supported version in security.adoc

Felk:
This seems to be outdated, since no more 1.x releases were provided after 2.x was released


Felk:
btw I noticed this while answering a question on stackoverflow: [Where to find information about Quarkus supported versions?](https://stackoverflow.com/q/69943423/3688648)

gsmet (approved):
Good catch, thanks!

Felk:
I noticed that this change has effectively been reverted by https://github.com/quarkusio/quarkusio.github.io/pull/1242 because that pull request introduced duplicate markdown files with outdated content. Was that intended?

PR #1212, opened 2021-11-17
Move Content Security Policy back to header in meta.

insectengine:
Adding the CSP to the header and removing it from the config.yml since that only works locally.


maxandersen (commented):
ha - good catch!

PR #1287, opened 2022-01-06
Update to new roadmap GitHub project, while not finished it's vastly more uptodate than the old one :)

emmanuelbernard:


gsmet (approved):

maxandersen:
the new url is not publically viewable yet - that's what I needed to fix first. will move it out in public and update this if need be.

PR #1288, opened 2022-01-07
Correct typo

ynojima:


gsmet (approved):
Thanks! I also pushed the fix to `_data/guides-main.yaml`

PR #1290, opened 2022-01-08
Fix typo in about.html that led to 404

alexisgeoffrey:
Changes link from `/Kubernetes-native/` to `/kubernetes-native/`


gsmet (approved):
Good catch, thanks!

PR #1293, opened 2022-01-12
initial attempt on tabbed code blocks

maxandersen:
basic attempt on adding tabbed blocks - but for some reason it freezes all my browsers when trying to load it :/

couldn't spot what is causing it.


gsmet:
Will push some examples here once I'm done with my experiments in the main repo.

_guides/includes/create-app.adoc
```diff
@@ -0,0 +1,60 @@
+[role=primary]
+.CLI - Maven project
```
kenyee, line 2: This wraps weirdly on a phone FYI...

assets/javascript/codeBlockSwitch.js
```diff
@@ -0,0 +1,52 @@
+function addBlockSwitches() {
+    $('.listingblock.primary, .sidebarblock.primary').each(function() {
+        primary = $(this);
```
Ladicek, line 3: Nitpicking (but not really): this looks like you want to declare a local variable, but you're actually unintentionally making `primary` a global variable (this is also known as polluting the global scope). If you really want to declare a local variable (and it totally looks like that's the case here and elsewhere in this script), use `let` (or `var`, but `let` has the civilized meaning).

It's enough that all the functions declared in this script go into the global scope :-) (It seems to me that they could easily be declared inside the `$(document).ready` callback and everything would work just fine.)

gsmet: FWIW, this is not my code :). I kept it as close as possible as upstream and fixed it to make it work with jQuery and the other things I needed.

gsmet: @Ladicek btw, if you want to fix things, be my guest. I think we will need to tweak this thing a lot so we might as well improve it.

Ladicek: Fair enough :-)

Ladicek: Pushed a commit with this change. All functions are now declared within the on-ready callback, and all local variables are actually local variables.

_guides/includes/create-app.adoc
gsmet: @kenyee thanks for the heads up. @insectengine will tweak this thing to make it prettier and more compatible.

gsmet:
I'm closing this one as James prefers a branch in the main repo to collaborate.

PR #1310, opened 2022-01-25
Fix copy to clipboard and deploy it everywhere

gsmet:


gastaldi:
Looks Awesome! Here are my suggestions, which can be done in a separate PR if you agree: 

- Display the Clipboard icon only when hovering the snippet area
- Change the clipboard icon to a check ✔️ to indicate that the copy succeeded

gsmet:
I did both.

The check icon is a cool idea. I wasn't happy about the result and I struggled with some tooltip libraries but in the end it wasn't really worth it to include a tooltip library just for that. But with the icon, it's perfect.

assets/javascript/copy.js
```diff
@@ -19,3 +19,13 @@ codes.forEach((code) => {
...
 let clipboard = new ClipboardJS('.btn-copy');
+clipboard.on('success', function(e) {
+    e.clearSelection();
+    e.trigger.className = e.trigger.className.replace("fa-clipboard", "fa-check");
+    e.trigger.setAttribute("title", "Copied!")
+
+    setTimeout(function() {
+        e.trigger.className = e.trigger.className.replace("fa-check", "fa-clipboard");
+        e.trigger.setAttribute("title", "Copy to clipboard")
+    }, 1000);
```
gastaldi, line 30: I think this can wait a bit longer (maybe 3s)?

gastaldi (approved):
Excellent! Added a minor suggestion

maxandersen:
This is awesome.  Works on mobile too. At least safari iOS :)

assets/javascript/copy.js
gsmet: aaaah, GH showed your comment when I clicked the Merge button.

gsmet: I pushed an update with 2 seconds, I think it's a good compromise.

PR #1315, opened 2022-01-28
Add Google Cloud services guides

loicmathieu:
@gsmet

I added a single entry to the Cloud section to avoid having too many guides on it. AWS services was added with one entry per service, maybe I should do the same ?

I added the secret manager guide under the miscelaneous section (as was done for Consul), I'm not sure it's the best place but there is no Configuration section and the core one is already packed with a lot of guide.

### DIFF
--- a/_data/guides-main.yaml
+++ b/_data/guides-main.yaml
@@ -228,6 +228,26 @@ categories:
         url: https://quarkiverse.github.io/quarkiverse-docs/quarkus-amazon-services/dev/amazon-s3.html
         description: This guide covers how to use the Amazon S3 cloud storage in Quarkus.
         origin: quarkiverse-hub
+      - title: Google Cloud BigQuery
+        url: https://quarkiverse.github.io/quarkiverse-docs/quarkus-google-cloud-services/main/bigquery.html
+        description: This guide covers how to use GCP BigQuery in Quarkus.
+        origin: quarkiverse-hub
+      - title: Google Cloud Bigtable
+        url: https://quarkiverse.github.io/quarkiverse-docs/quarkus-google-cloud-services/main/bigtable.html
+        description: This guide covers how to use GCP Bigtable in Quarkus.
+        origin: quarkiverse-hub
+      - title: Google Cloud Firestore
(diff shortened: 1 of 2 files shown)


_data/guides-main.yaml
```diff
@@ -234,6 +234,26 @@ categories:
...
+        url: https://quarkiverse.github.io/quarkiverse-docs/quarkus-google-cloud-services/main/firestore.html
+        description: This guide covers how to use GCP Firestore in Quarkus.
+        origin: quarkiverse-hub
+      - title: Google Cloud Spanner
+        url: https://quarkiverse.github.io/quarkiverse-docs/quarkus-google-cloud-services/main/spanner.html
+        description: This guide covers how to use GCP Spanner in Quarkus.
+        origin: quarkiverse-hub
+      - title: Google Cloud Storage
+        url: https://quarkiverse.github.io/quarkiverse-docs/quarkus-google-cloud-services/main/storage.html
+        description: This guide covers how to use GCP Spanner in Quarkus.
```
gsmet, line 255:
```suggestion
        description: This guide covers how to use GCP Storage in Quarkus.
```
?

loicmathieu:
@gsmet I applied your suggestions.

Maybe I miss something but testing the preview I cannot find the added guides.
I didn't add anything inside the `_redirects` directory as I thought it was for redirection, maybe it's required ?

gsmet:
You added them to `main` so they are only shown if you switch to the `main` version. You need to sync it with `-latest` (I think the two files are in sync right now so just try to copy main to latest)

loicmathieu:
By doing a copy/paste there are a few space differencies (better to fix them) and the hibernate reactive guide is missing.
I think it's good to add it there also so I'm commiting my copy/paster.

gsmet:
OK, I'll wait for 2.7 to be released to merge it as the Hibernate Reactive guide is coming with it.

Thanks!

loicmathieu:
Great, I'll also include in 2.7 the 1.0.0 of the Google Cloud services extensions as I decided to promote to 1.0 the version to denote it as stable, so it's better to wait for 2.7 anyway ;)

PR #1350, opened 2022-03-02
Add a button to the Guides page in the home screen

gastaldi:
This makes much easier to navigate to the guides page


gsmet:
It's a good idea but I will let @insectengine decide as it touches the home page.

insectengine (approved):

PR #1406, opened 2022-04-24
Add tooltip inline macro in order to display extra info in config table

michalvavrik:
Enum constant documentation is not propagated to Quarkus website. One example- `io.quarkus.oidc.common.runtime.OidcCommonConfig.Tls.Verification` has clear documentation of what `REQUIRED`, `CERTIFICATE_VALIDATION` and `NONE` means, however you won't find that in documentation. The meaning of this constants was enigma for me till I checked the codes, but if we want developers to do that, there is no point of providing Quarkus website at the first place.

My idea is that we use tooltip to show description. In `io.quarkus.annotation.processor.generate_doc.ConfigDoItemFinder#extractEnumValues` if there is constant description, we append value like this `tooltip:REQUIRED[+++Certificates are validated and hostname verification is enabled. This is the default value.+++]` and we alter `io.quarkus.annotation.processor.generate_doc.SummaryTableDocFormatter#TABLE_ROW_FORMAT` to prefix respective cell delimiter with `a`. 

tooltip-hover

I am happy to implement the other part (in quarkus annotation processor), but it does not make sense till we have a way to display description.


michalvavrik:
I just mentioned that in some cases you use `border-radius: 10px;`, I can change my commit to match that if that's preferred.

insectengine:
I'm fine with the new tooltip (with a 10px radius to match our buttons and cards).

michalvavrik:
thanks @insectengine , border radius fix is pushed

gsmet (approved):

gsmet:
This is a good idea for sure. Let's see what CI has to say.

gsmet:
Merged, thanks!

PR #1426, opened 2022-05-11
old faq/security md and its includes/layouts removed to not overshadow the correct .adoc

maxandersen:
user reported question about quarkus.io/security only listed 1.x.

Turns out the [big web 2.0 update commit](https://github.com/quarkusio/quarkusio.github.io/commit/8940e387f0570e1c9d0fdf8d782eb1b920600676) in December overwrote the [commit](https://github.com/quarkusio/quarkusio.github.io/commit/b31a29a84d4fcf38567c1a62371730aec0675089) that moved faq & security to use adoc instead of markdown.


gsmet (approved):
I checked and it looks good, thanks!

PR #1434, opened 2022-05-16
update language selector for usability and official site designation.

insectengine:
Changes to the language pulldown

1) for better usability, changed "Japanese" to be legible in Japanese characters in header and footer.
2) In primary navigation, changed "English" to "Official (English)" to help the indicate to the user the English version of the site is the main project site.
3) Added thin banner on home page between the Navigation and the hero. This banner uses a yml (languages.yaml) file to store the content for easier translation. The content reads; "The English version of Quarkus.io is the official project site. Translated sites are community supported on a best-effort basis."  We want to use this disclaimer so the users understand that the content of translations may lag behind the English site. One thought would be to make that banner not visible for the English site but visible and translated for any non-english site. 

Thoughts on making the banner only visible on the translated sites?

### DIFF
--- a/_data/language.yaml
+++ b/_data/language.yaml
@@ -1,1 +1,0 @@
-headline: "The English version of Quarkus.io is the official project site. Translated sites are community supported on a best-effort basis."
--- a/_data/language.yaml
+++ b/_data/language.yaml
@@ -1,0 +1,1 @@
+headline: "The English version of Quarkus.io is the official project site. Translated sites are community supported on a best-effort basis."
--- a/_data/languages.yaml
+++ b/_data/languages.yaml
@@ -1,2 +1,2 @@
 language:
-  - headline: "The English version of Quarkus.io is the official project site. Translated sites are community supported on a best-effort basis."
+  - headline: 'The <a href="https://quarkus.io">English version of quarkus.io</a> is the official project site. Translated sites are community supported on a best-effort basis.'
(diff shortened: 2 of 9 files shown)


insectengine:
@gsmet  - I agree. I just don't know how to programmatically make it happen. Any suggestions on how to achieve this?

maxandersen:
the content of CNAME is different dependent on the site being rendered.

looking into see if can load that while rendering and have jekyll react.

maxandersen:
@insectengine I've rebased and pushed change so please update/pull your branch before doing changes.

i've added pluging that loads content of CNAME into `site.cname` and then you can check what build is happening looking at site.cname. Thus i added a `if site.cname!="quarkus.io"` to make it only show up non l10n sites.

note - i noticed it is only setup to include on front page - should it not be on all pages ?

gsmet:
The website should be defined here in the various websites: https://github.com/quarkusio/quarkusio.github.io/blob/develop/_config.yml#L21 .

So it should just be a matter of testing `site.url`.

maxandersen:
I did consider using site.url but i wasn't 100% sure we didn't rely on that actually being the official url in places and it requires adding config override in every xx.quarkus.io repo.

by using CNAME it works with current setup.

when i'm back i do wanna refactor the xx.quarkus.io setup to have a shared place so we can do such adjustements in one place rather than in every site...but thats for *after* my pto ;)

PR #1463, opened 2022-06-20
remove double l10n banner on frontpage

maxandersen:
without this cn.quarkus.io and ja.quarkus.io has a double mention of localized vs official site.

only the top level one should be visible/included.


maxandersen:
if you wanna test it locally just remove CNAME file before you call jekyll.

gastaldi (approved):

PR #1505, opened 2022-08-05
provide faster serving for blog posts/website updates

maxandersen:
this adds a separate config that does nothing but just exclude guides/version and then provide command to use it.

difference on my machine is 5 seconds instead of 60+ seconds for full render.


gsmet:
Cool!

PR #1510, opened 2022-08-11
Make it easy to copy environment variables in config tables - add button

michalvavrik:
Configuration tables contain environment variables that are hard to copy as when you click on them, the property description collapse and the env variable is gone. @famod suggested we should do something about that here https://github.com/quarkusio/quarkus/pull/26408#issuecomment-1198304829

This PR adds a copy button right behind each such environment variable and also prevents collapsing.

image

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -64,6 +64,7 @@ asciidoctor:
     sectanchors: ''
     icons: font
     outfilesuffix: ''
+    add-copy-button-to-env-var: true
 
 # Pages permalink
 defaults:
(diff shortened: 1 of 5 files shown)


michalvavrik:
@insectengine WDYT?

michalvavrik:
@gsmet please assign reviewer

insectengine:
Sorry... Don't know how I missed this. This looks good to me. @gsmet

gsmet:
@michalvavrik have you checked the page with ALL the configuration properties? I know it's very sensitive as it contains a lot of stuff and even a small amount of Javascript can make it dangerously slow.

michalvavrik:
I tried treeprocessor and parser but it's too slow, therefore the PR now uses inline macro and avoid adding the button via JS.

machi1990 (approved):

gsmet (approved):

gsmet:
@michalvavrik thanks! (better late than never ;))

PR #1519, opened 2022-08-18
Update URL on get-started-code-steps.html

IgnasiCR:
This URL doesn't redirect to the page and confuse the user.


gsmet (approved):
Thanks!

PR #1522, opened 2022-08-23
add sticker pdf files and page

insectengine:
Adding PDF files for Quarkus stickers (the logos, icon and "rocking dukes") and page for viewing while NOT including a direct link in the primary or footer navigation. The purpose for these to be on the site but not readily visible so for us to be able to share them with specific community members we want to share them with.


gsmet (changes requested):
In the preview, the image for `Rocking Duke - Bass 1` is missing. Could you check that everything is in order?

~~Also is it normal this page is not linked anywhere?~~ Scratch that, I should have read the description :).

BTW, I had an observation about the first two with the Quarkus logo: I wonder if it would be better with a less convoluted shape e.g. with a more rectangular shape (with rounded corners probably) rather than following the letters.

gsmet (approved):
Thanks!

PR #1525, opened 2022-08-26
Solving problems with extensions 2/n

gsmet:


_posts/2022-08-30-solving-problems-with-extensions-2.adoc
```diff
@@ -0,0 +1,193 @@
...
+    }
+}
+----
+<1> Our class implements `AnnotationsTransformer`.
+<2> We inject the Jandex index in our transformer as we will need it to detect composition.
+<3> We are only interested in fields so let's apply our transformer to fields only.
+<4> If the field is not annotated with `@Inject`, it is of no interest to us.
+<5> If the field type is `GlobalMetadata`, `GroupMetadata` or `CommandMetadata`, we know it is the responsibility of Airline to inject it.
+<6> We are also detecting composition.
+<7> We remove the `@Inject` annotation in this case. Make sure you don't forget to finalize the transformation with `.done()`.
```
yrodiere, line 170: I was surprised at first that this wouldn't break Airline injection, but of course that's because we don't actually remove the annotation from the class, we just remove it from the index used by Arc. Maybe that should be stressed out?

yrodiere (approved):
Looks great! Just one minor comment.

_posts/2022-08-30-solving-problems-with-extensions-2.adoc
gsmet: Yeah, it was in the schema but I added some text to make it extra clear.

gsmet: Thanks for the review!

PR #1458, opened 2022-06-09
Documentation build - Add new Docs main page and navigation.

insectengine:
Created new Documentation main page to replace the current guides page. This page does not have the appropriate coding to pull new guides and build properly yet. I'm creating this PR in hopes that someone on the larger team has the skillset to compete the necessary hooks to make it build properly.

I've also replaced the guides page link in the primary navigation with the new Documentation page.

PR #1566, opened 2022-11-03
Website updates for diataxis framework content

ebullient:
Supersedes #1458 (Combines style/layout with required code changes for iteration, etc.)
Website half of https://github.com/quarkusio/quarkus/issues/28484

I'll keep this up to date, will need to tweak styles and behavior as https://github.com/quarkusio/quarkus/pull/29022 is integrated.


insectengine (approved):
Works locally for me.

_includes/index-doc-item.html
```diff
@@ -0,0 +1,17 @@
...
+{% else %}
+  {% assign guide_url = site.baseurl | append: '/version/' | append: include.docversion | append: include.url %}
+{% endif %}
+<div class="{% if include.origin == 'quarkiverse-hub' %}quarkiverse {% endif %}grid__item docs-card {{ include.class }}">
+  <h4><a href="{{ guide_url }}" {% if is_external_guide %}target="_blank" {% endif %}>{{ include.title }}</a></h4>
+  <div class="description">{{ include.summary | markdownify }}</div>
+  <div class="keywords">{{ include.keywords }}</div>
+  <div class="categories">{{ include.categories }}</div>
+  {% if include.origin == 'quarkiverse-hub' -%}
+  <div class="origin">Quarkiverse Hub</div>
```
maxandersen, line 15: is there a reason this is calling out Quarkiverse Hub?

maxandersen (approved):
general approach is introducing separate versioned guide layout keeping the existing in place thus looks good to merge IMO.

The new layout renders a bit weird in my eyes and I don't grok the "quarkiverse hub" reference but those are things we can adjust as we iterate IMO.

_includes/index-doc-item.html
ebullient: because the old code did.

ebullient: if/when we want to style additional origins, we can remove the condition on including the div.

maxandersen: ha - it hadn't noticed we do that  as it shows up much further down. 
its just that I don't grok why its relevant to mark it with quarkiverse hub on docs ... If anything I wish we had docs per platform releases instead of having what looks like mix of in/outside platform docs.

ebullient: we could remove the quarkiverse docs at this point... we have Holly's site now..

maxandersen: I agrée we should clean this up somehow but just removing quarkiverse hub docs does not feel like the right one as the platform has docs that are from quarkiverse. 

We should figure out what the docs site is trying to be? Docs for the core or platform release? Right now we have docs from parts of core and platform but not all and all config guide is only core.

PR #1585, opened 2022-11-30
Fix typos

coproduct:
Found these while I was browsing the guides. I hope this is the correct repo to submit to, I didn't find the relevant sentences in the main one.


gsmet (approved):
Thanks!

PR #1593, opened 2022-12-14
Cross-link to release documentation

holly-cummins:
I had a hard time finding https://github.com/quarkiverse/quarkiverse/wiki/Release by google search. I'm hoping if we link to it, it will be a bit more-indexed by google, and also easier to navigate to from the pages google does find.


maxandersen:
google does not index github wikis. would be interesting if we could put this in normal docs so it could be indexed.

holly-cummins:
I always suspected Google wasn't indexing the wiki, but then decided that surely it must be, and I was searching wrong. I honestly can't see much value in having the wiki if Google doesn't index it - the information is basically invisible, unless we share a direct link to it for someone, or someone is *very* enterprising in manually navigating. 

I like @maxandersen's suggestion of just putting this stuff in the normal docs so it's findable. Since the publication process *feels* like similar levels of friction for the GitHub wiki and the normal docs site (both need a change in a GitHub repo and then a PR), I wonder what value we're getting from the wiki. I guess maybe there's extra layers of review/extra expectation of minimum standards/extra deploy delay in the normal wiki site?   

(It also explains why I could never find my Quarkus 3 cheat sheet by searching, and always had to go back and find an email where I'd sent it to someone to get the link.)

holly-cummins:
Here's a link about the indexing of GH wikis: https://github.com/isaacs/github/issues/1683

holly-cummins:
I've raised https://github.com/quarkiverse/quarkiverse/issues/93 to discuss the indexing of the wiki and workarounds, more generally.

gastaldi (approved):

PR #1608, opened 2023-01-17
OpenTelemetry - Fix missing char

aolle:


maxandersen:
thanks for the fixes - you should submit these on quarkusio main repo as that is used for the source of the docs.

are you able to do that?

aolle:
Sure, I am going to send the PR on that repo. Thank you!

PR #1619, opened 2023-01-25
Correct vertical alignment of navbar on mobile

### LINKED ISSUE #1618: Fix alignment of primary navigation (logo and hamburger icon)
The logo and "hamburger" more icon have vertical alignment issues on smaller screen sizes.

holly-cummins:
Resolves #1618 

I’ve switched everything (in this area) to flex box. This allowed me to remove various bits of vertical padding that I think were being used to try and get things into position. This slightly upsetting graphic shows the before and after overlaid, with the vertical shift:

image

On normal screen, things are unchanged: 

image

I had to do a bit of work to get the the logo and menu to wrap at the same screen width, but I’ve now got the same wrapping behaviour as before. I’m using flex gutters to set the spacing between the elements, which I think is cleaner. 

On medium sized screens, where the logo and nav menu stack, the Quarkus logo shifted up by 10 pixels, and the nav bar is 10 pixels smaller. I think this is ok. (Before, everything was positioned a bit below centre.)

image

There’s a bit of ugliness where the vertical padding on the `main-nav` bar is different on mobile and desktop in order to preserve the previous appearance, but I think it’s ok. If we want to tidy it, we could just get rid of one of the special cases and accept the slight change in navbar dimensions. 

I also had to use a `!important` to get wrapped items into the menu, since I didn’t want to risk re-ordering the different @media queries.

Disappointingly, the sass didn’t actually end up shorter, once I’d covered all three cases. It’s hopefully more robust, although there is some fragility still in how the media queries cascade.

### DIFF
--- a/_sass/core/includes/header-navigation.scss
+++ b/_sass/core/includes/header-navigation.scss
@@ -19,7 +19,7 @@ $nav-height: 70px;
   display: inline;
 }
 
-// Navigation 
+// Navigation
 nav {
   float: right;
   padding-top: 1rem;
@@ -115,8 +115,7 @@ nav ul.nav-list {
     z-index: 1;
   }
   .logo-wrapper {
-    display: block;
-    margin: 0 auto;
+    display: flex;
     padding-top: 0;
   }
   nav {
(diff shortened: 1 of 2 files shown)


insectengine:
Ran this locally and it looks good to me.

PR #1628, opened 2023-01-31
Remove horizontal padding around Quarkus logo

holly-cummins:
I spotted that some padding on the Quarkus logo means it's not aligned with other content on the left margin. 

image

I've discussed with @insectengine and we don't think there's any reason to bump the logo right, so I've removed the padding.


insectengine:
Looks good to me.

PR #1630, opened 2023-02-01
Spacing tweaks to improve reproducibility of navigation header

holly-cummins:
I’ve continued the flexbox-ification of the navigation bar, which should make the layout more maintainable in future. I aimed for basically the same as now, but with a tolerance of a few pixels where the code would be cleaner. 

In particular: 
- My changes reduce the height of the navbar by about 2-3px. Let me know if we want it back, and I can adjust some padding to get it back.
- On mobile, the spacing between the elements got a bit more squished, to 56px from 59px. Again, if we think it’s important I can adjust paddings to get it back, although a round number like 58 or 60 might be best. 
- 
I was able to get rid of a few !importants, which is always a good feeling. (We were using flex on mid-size screens, and block on large screens.)

For the menu chevrons, I switched to using padding to control the spacing, rather than a space character. This gives us more flexibility. 

I was hoping I'd reduce the number of lines of css, and I don't think I did, really, but I'm hoping these are still good changes. They have allowed me to get the extensions sub-site headers 100% aligned with the main site, except for the icons (and see https://github.com/quarkusio/extensions.quarkus.io/issues/87).

### DIFF
--- a/_includes/header-navigation.html
+++ b/_includes/header-navigation.html
@@ -11,7 +11,7 @@
     </label>
     <ul id="menu" class="menu">
       <li class="dropdown">
-        <span href="{{site.baseurl}}/about/">About <i class="fas fa-chevron-down"></i></span>
+        <span href="{{site.baseurl}}/about/">About<i class="fas fa-chevron-down"></i></span>
         <ul class="submenu">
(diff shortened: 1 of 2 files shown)


holly-cummins:
Here are some overlays of old vs new. 

Desktop: 

image

holly-cummins:
Medium width: 
image

holly-cummins:
Mobile. You can see the different padding on the menu here (the new one is less spaced):

image

image

insectengine (approved):

PR #1634, opened 2023-02-06
Include 'Extensions' submenu on site header

holly-cummins:
This is it! The go-live for http://quarkus.io/extensions. FAO @maxandersen and @insectengine .

The extensions pages definitely are [not perfect](https://github.com/quarkusio/extensions/issues), and not ‘finished’. However, I believe at this point the site is at the stage where it has enough value that it’s better to link to it than to not link to it. 

This request should not be merged until https://github.com/quarkusio/quarkusio.github.io/pull/1630 is merged, or there will be lots of spacing jumps when the ‘browse extensions’ link is clicked. The one other issue I might consider a blocker to cross-linking is https://github.com/quarkusio/extensions/issues/87, which causes tiny ‘jump’ on the menu chevrons when the ‘browse extensions’ link is clicked. I think it's minor enough we don't need to block on it, but I'm happy to be persuaded otherwise.

The best link for the 'using extensions' item on the menu is a fragment within the FAQ page, and the active highlighting doesn't work for it. The root problem is we don't have a great page to link to, so we should do https://github.com/quarkusio/quarkus/issues/27946, and then that fixes the problem.

I’ve removed `header-navigation-full.html`, since nothing was using it (based on search, and observation). Once I’d added the extension content to the main header-navigation, the two pages were almost identical. The ‘full’ page was actually less complete, since it was missing the translations menu. (I assume because we stopped double-maintaining it at some point.)


maxandersen:
in the launch I think it is important we add some info on the page to educate people if they find something is missing report it (with link to issue tracker) and if they find their extension is not represented correctly or simply missing what they should do.

holly-cummins:
Ok, so also block on https://github.com/quarkusio/extensions/issues/71?

maxandersen:
also - still seem to see 3.0.0.Alpha as the recommended/try version?

holly-cummins:
> also - still seem to see 3.0.0.Alpha as the recommended/try version?

Yep, this is https://github.com/quarkusio/extensions/issues/53. I think the question isn't "is this unfortunate, do we want to fix it?" (yes, and yes, obviously we do, and as soon as possible) it's "is there enough value in the resource we have that we should let people start using it?". 

I've created a label 'blocks-go-live' that we can use to highlight the issues that we think prevent us sharing the extensions link: https://github.com/quarkusio/extensions/issues?q=is%3Aopen+is%3Aissue+label%3Ablock-go-live

gsmet:
I spotted several issues. Maybe let's have a look at these before we decide to merge this. I will create some issues in the extensions repo.

holly-cummins:
> yeah; before we make itfully public.
> 
> the missing camel/kogito is problematic too - should fix it or at least state its missing.

Getting the other platforms into the registry is covered by https://github.com/quarkusio/registry.quarkus.io/pull/92. That pull request has been outstanding for about six months, so I'll raise a issue to document the limitation, and mark it as blocking go-live: https://github.com/quarkusio/extensions/issues/100

holly-cummins:
> I spotted several issues. Maybe let's have a look at these before we decide to merge this. I will create some issues in the extensions repo.

Thanks. Could you label them with the `block-go-live` label?

holly-cummins:
Another 'bar' that we can compare these pages to is https://quarkus.pro/extensions/. At the moment, that's what's available to users to explore extensions. Do we think what we have at the moment is perfect? Goodness, no. Do we think there are lots of things to fix? Absolutely, yes. Is it a better option for users than https://quarkus.pro/extensions/? That's what we'll need to decide.

Would users be better off knowing about this page than not? Well, if it's tooo bad, no - there's reputation damage if we publish something which is has too many deficiencies. On the other hand, there's also reputation damage in not making the information available at all, so it's a balancing act.

maxandersen:
its more code.quarkus.io we should compare to - and here camel are listed.

I fully get we want it out but a few of these will generate more questions and "WTH?" (i.e. why is camel not listed here - is it somehow secondary citizen?) than we would want to generate.

the quarkus 3 alpha use I'm less worried about - especially if added a "UNDER DEVELOPMENT/known issues" or something on the page :)

holly-cummins:
I (and @gastaldi) have resolved everything [labelled 'block-go-live'](https://github.com/quarkusio/extensions/issues?q=is%3Aopen+is%3Aissue+label%3Ablock-go-live) on the extensions side, so I think we should have another look at this. I've rebased so that we get a new preview, to get the feel for navigation between the two sites.

https://quarkus-io-pr-1634-preview.surge.sh/ has the surge preview 

The things to look at are: 
- Is the navigation back and forth smooth enough? I have a few styling and performance issues in this area that I want to look at, but hopefully not blockers. 
- Are we happy with the other links in the menu? @insectengine and Jeff are looking at some content generation for this, but I think the current endpoints work as 'better-than-nothing' content in the interim.
- Is https://quarkus.io/extensions/ good enough? (With the context that I'm continuing to make improvements.)

gastaldi:
@holly-cummins the Quinoa extension is looking weird. Can we abbreviate longer descriptions to avoid messing with the UI ? 
 
image

gastaldi:
Also extensions with `Parent POM for Quarkiverse projects that includes the default release and artifact publishing related configuration` are missing their description (this is inherited from `quarkiverse-parent`)

holly-cummins:
> @holly-cummins the Quinoa extension is looking weird. Can we abbreviate long descriptions to avoid messing with the UI? Something like https://css-tricks.com/snippets/css/truncate-string-with-ellipsis/
> 
> image

Yeah, this is https://github.com/quarkusio/extensions/issues/25 and https://github.com/quarkusio/extensions/issues/97 (it's sort of covered by both).

holly-cummins:
> Also extensions with `Parent POM for Quarkiverse projects that includes the default release and artifact publishing related configuration` are missing their description (this is inherited from `quarkiverse-parent`)

Ooh, do you mean ones like https://quarkus.io/extensions/io.quarkiverse.jdbi/quarkus-jdbi? I'd wondered what was going on with those and thought it was a copy and paste. So do you think in the UI we shouldn't show what the registry gives us, but instead detect that it's kind of a silly description and blank it out? 10 projects which have that description. 

We might also want to make a change in the maven extension plugin to not put `Parent POM for Quarkiverse projects that includes the default release and artifact publishing related configuration` into the registry in the first place (is that what's doing it?). I guess sometimes inheritance makes sense but sometimes it doesn't, which makes it harder to decide when not to do it.

gastaldi:
> We might also want to make a change in the maven extension plugin to not put `Parent POM for Quarkiverse projects that includes the default release and artifact publishing related configuration` into the registry in the first place (is that what's doing it?)

Exactly.

holly-cummins:
Thanks, @gastaldi. I've raised https://github.com/quarkusio/quarkus/issues/31634 for the longer term fix, and https://github.com/quarkusio/extensions/pull/149 for the tactical suppression.

maxandersen (commented):
was reviewing/testing to merge in but found issues - marked them so they show up on https://github.com/quarkusio/extensions/issues?q=is%3Aopen+is%3Aissue+label%3Ablock-go-live 

lets review that and see if fixable or liveable.

PR #1643, opened 2023-02-14
Update cli-tooling.adoc

smartinus44:
Adding more details to install JBang on RHEL 9.x system if a dependency is missing.

### DIFF
--- a/_guides/cli-tooling.adoc
+++ b/_guides/cli-tooling.adoc
@@ -92,6 +92,15 @@ have a JBang version older than v0.56.0 installed. Please remove or upgrade it t
 
 If you are installing JBang for the first time, start a new session to update your `PATH`.
 ====
+[CAUTION]
+.Dependency required on RHEL 9.x
+====
+In order to install JBang, you need to install the `tar` command on your system.
+[source,bash]
+----
+dnf install tar
+----
+====
 ****
 
 [role="secondary asciidoc-tabs-sync-sdkman"]


_guides/cli-tooling.adoc
```diff
@@ -92,6 +92,15 @@ have a JBang version older than v0.56.0 installed. Please remove or upgrade it t
 
 If you are installing JBang for the first time, start a new session to update your `PATH`.
 ====
+[CAUTION]
+.Dependency required on RHEL 9.x
+====
+In order to install JBang, you need to install the `tar` command on your system.
+[source,bash]
+----
+dnf install tar
```
gsmet, line 101: @maxandersen I wonder if your installer should use another command if `tar` is not available. But I don't know what is actually available on a default install of RHEL 9.

smartinus44: The context: I was using RHEL 9.1 on a local virtual machine installed with a DVD iso of RHEL.

maxandersen: A Linux without tar...what have the world become :)

maxandersen: Btw. I would rather fix the install to inform user it needs tar rather than repeating it in docs everywhere. 

What error message did you get?

maxandersen: Actually in worst case if user already have java we can use jar to do the unzip. Worth a shot.

smartinus44: it was an error saying to install tar, once i installed it there was no problems anymore...

PR #1660, opened 2023-03-15
add socials page

insectengine:
This creates a single local page for us to list all our social links. We'll then be able to create a QR code for this page and display it at conferences and in presentations. The page itself is set up for viewing on mobile use (big easy buttons and no distracting content).


insectengine:
Connect with us on social media - Quarkus

Preview on mobile

PR #1663, opened 2023-03-22
Update and rename security.adoc to security-policy.adoc

rolfedh:
Fixes: https://github.com/quarkusio/quarkusio.github.io/issues/1664


rolfedh:
See https://github.com/quarkusio/quarkusio.github.io/pull/1665 instead.

Issue #1664, opened 2023-03-22
Update Security Policy document to improve `Quarkus Security` search results


rolfedh:
**Describe the enhancement**

When we had the security overview concept in `security.adoc`, searching in Google for `Quarkus Security` would link to this document, with the `Security Policy` document not really very visible as far as I recall.

Now, if I search in Google for `Quarkus Security` I see `Security - Quarkus` as a 4th candidate pressing on its link brings me to https://quarkus.io/security/ which starts with a message like `We take security bugs seriously` - while I wanted to get to the security overview instead - which is still shown as the first candidate but seeing `Security - Quarkus` not linking to the overall overview concept is confusing.

 

Proposal: Rename the Security Policy doc to precisely represent what it is, `security-policy.adoc` and the title of the document should also be renamed from `Security` to `Security Policy`.

The `Security Policy` (name correctly here) link at the bottom of  https://quarkus.io/ should be fixed to link to `security-policy.adoc`

**Which pages**
(add links or names for the affected pages)

rolfedh:
Also see https://issues.redhat.com/browse/QDOCS-215

rolfedh:
@sberyozkin Although we discussed this issue in https://issues.redhat.com/browse/QDOCS-215, would you mind restating your original concern or support for this issue here for the benefit of @gsmet @maxandersen @insectengine?

PR #1665, opened 2023-03-22
Update and rename security.adoc to security-policy.adoc

rolfedh:
Fixes: https://github.com/quarkusio/quarkusio.github.io/issues/1664

### DIFF
--- a/security.adoc
+++ b/security.adoc
@@ -1,12 +1,12 @@
 ---
 layout: base-fullwidthcontent
-title: Security
+title: Security policy
 subtitle: The Quarkus team and community take all security bugs very seriously. You can find our guidelines here regarding our policy and security disclosure.
 permalink: /security/
 ---
 
 ////
-    Content copied from https://github.com/quarkusio/quarkus/blob/master/SECURITY.md 
+    Content copied from https://github.com/quarkusio/quarkus/blob/master/SECURITY.md
     Edit there too
 ////
 


rolfedh:
@maxandersen @sberyozkin PTAL

rolfedh:
@gsmet Please review and merge.
cc: @sberyozkin

rolfedh:
@gsmet Updated and ready to merge.
cc: @sberyozkin

sberyozkin:
Hi @gsmet Can you please look at this PR ?

gsmet (changes requested):
I'm really not sure this change will change anything for Google searchability of the security guide... and it's breaking any external link to our security page which is a no-no.

We at the minimum need to put a redirect in place.

But even so, I'm not especially excited about it.

@maxandersen @insectengine WDYT?

rolfedh:
> We at the minimum need to put a redirect in place.

Thanks. I've just added `_redirects/security.md` to address this.
I've also asked @sberyozkin to provide support for or a fuller explanation of the original issue in https://github.com/quarkusio/quarkusio.github.io/issues/1664

maxandersen:
I've read but still I don't grok why changing this url is supposed to help? 

It is a good thing it's just quarkus.io/security. 

Why require the longer harder to remember/copy/read quarkus.io/security-policy url?

maxandersen:
Changing the title I can buy or at least add security policy in top part of message but the url and redirect is a -1 from me.

rolfedh:
> I've read but still I don't grok why changing this url is supposed to help?

An [SEO best practice](https://www.google.com/search?q=seo+file+names&oq=seo+file+name&gs_lcrp=EgZjaHJvbWUqCQgAEEUYOxiABDIJCAAQRRg7GIAEMgYIARBFGDkyCAgCEAAYFhgeMgoIAxAAGA8YFhgeMggIBBAAGBYYHjIICAUQABgWGB4yCAgGEAAYFhgeMgYIBxBFGEDSAQg2MDMwajBqN6gCALACAA&sourceid=chrome&ie=UTF-8) is to align the file name and the title of the document.

image

maxandersen:
thats fine but lets please not use long winded "policy" names in URLs. URLs like these are to be remembered, typed and understood easily. 

quarkus.io/security is easy to remember, easy to type and most understand that page has stuff to do around quarkus security - let me read that .

quarkus.io/security-policy is not easy to remember, hard to type (is it with a or without a dash ? is it policy to not policy?, _ and if I see security-policy it is not a page I or many I can think of will go read :)z


and in addition quarkus.io/security is short and descriptive and use same keywords as in title and in page so I think we are covered.

rolfedh:
Absolutely, I'm happy we've discussed this and come to a consensus. My aim was to put into action what I thought Sergey was suggesting. As it stands, we've only agreed on revising the title. So, I'll remove the changes for the redirect, altering the file name, and updating the links. The PR should be ready to merge shortly.

rolfedh:
@gsmet ready to merge

rolfedh:
Actually `quarkusio/quarkus/blob/master/SECURITY.md` states that
`The canonical version of this document is hosted on the Quarkus website at [Quarkus security](https://quarkus.io/security/).` In other words, the file in this PR is the only one that needs to be updated.

rolfedh:
@gsmet The result of our conversations is a simple change to the title. I believe this is ready to merge now.

PR #1687, opened 2023-04-21
Fix guide version dropdown on All config page and Build Items page

michalvavrik:
Fixes guide version selection dropdown on All config page and Build Items page as it doesn't work due to typo in CSS ID selector. You can try that it doesn't work here https://quarkus.io/guides/all-config


michalvavrik:
Not sure what the PR preview problem is, but tested locally, works fine.

holly-cummins:
I had other PR previews fail around the same time, so it's not just this PR. From the logs, looks like maybe a DNS glitch?

PR #1728, opened 2023-06-15
Improve searchability of OIDC documents

sberyozkin:
This PR adds more key words to OIDC docs to improve their searchability.
Every OIDC doc gets extra `oidc`, `oauth2` tags because in reality Quarkus OIDC support both `oidc` and `oauth2`. `jwt` is also added since it is a base block of these technologies, so when users type `jwt`, they should be able to see OIDC docs too (in fact for now I added it to the OIDC services doc only which is probably enough).
Additionally the well-known providers doc gets an extra key word per every supported provider as discussed as F2F (CC @maxandersen )

### DIFF
--- a/_data/guides-latest.yaml
+++ b/_data/guides-latest.yaml
@@ -370,43 +370,43 @@ categories:
       - title: Using OpenID Connect (OIDC) to Protect Service Applications
         url: /guides/security-openid-connect
         description: This guide demonstrates how to use the OpenID Connect extension to protect your Quarkus JAX-RS service application using Bearer Token Authorization where the tokens are issued by OpenID Connect Providers such as Keycloak.
-        keywords: sso,jwt,security
+        keywords: sso,oidc,oauth2,jwt,security
       - title: Using OpenID Connect (OIDC) to Protect Web Applications
         url: /guides/security-openid-connect-web-authentication
         description: This guide demonstrates how to use the OpenID Connect extension to protect your Quarkus JAX-RS web application using the Authorization Code Flow and OpenID Connect Providers such as Keycloak.
-        keywords: sso,security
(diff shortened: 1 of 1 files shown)


michelle-purcell (approved):
@sberyozkin - Keywords LGTM :+1: 

Although, it looks like the current `data/guides-latest.yaml` is outdated as per my comments. 
@ebullient  - Does this file get dynamically updated by the doc builds? Should we be updating this manually every time we rename and redirect a Quarkus adoc file?


FYI: For SEO, where possible, we try to include the most popular keywords, phrases, or terms that customers might use to find stuff in the main title, headings, abstract, and main body of the content source. Google and other well-known search engines index more content elements than the Quarkus search function, which only seems to observe keywords/terms in the titles and abstract of the guide source plus other external files like the one above. See related GitHub issue #[33215](https://github.com/quarkusio/quarkus/issues/33215).

sberyozkin:
I tried to address your comments and got confused, because, right now:
* bearer and code flow authentication docs have
* bearer tutorial and ref guides docs have different file name structure
* impossible to find a code flow guide when landing on one of the bearer token docs
* impossible to find a bearer token tutorial when landing on the bearer token ref doc

I propose that in this PR we just take care of the tags/key words, as it is really a standalone PR, to address your comments properly we need to do some work first in the main repo.
Next, I'll open a PR to the main repo addressing the above 4 points and ask you to review
And then I'll open a PR here to address your comments above

Does it sound reasonable ? 
thanks

sberyozkin:
Let me merge this one as I've noticed Michelle approved, @michelle-purcell I'll follow up with some updates in the main repo as suggested above, and then I'll ping you to discuss how to best manage your suggestions above, thanks

PR #1733, opened 2023-06-27
add json file for Bruno

insectengine:

### DIFF
--- a/assets/json/03656937-19FD-4C83-9066-C76631D445EA.json
+++ b/assets/json/03656937-19FD-4C83-9066-C76631D445EA.json
@@ -1,0 +1,1 @@
+{"active":true,"deny_anonymous_ids":[],"deny_quarkus_versions":[],"refresh_interval":43200.000000000}


brunobat (approved):

gsmet:
@brunobat could you clarify what it is used for?

assets/json/03656937-19FD-4C83-9066-C76631D445EA.json
```diff
@@ -0,0 +1 @@
+{"active":true,"deny_anonymous_ids":[],"deny_quarkus_versions":[],"refresh_interval":43200.000000000}
```
maxandersen, line 1: should this not be enabled true?

gsmet (approved):
I will merge it to unblock things but I would like to understand why we need this file.

maxandersen:
> @brunobat could you clarify what it is used for?

its the file build analytics fetches to know if analytics is still active for the specific quarkus version.

i.e. imagine some bug causes data issues we can disable that specific version or some day we have to change the collection - then instead of old quarkus builds keeps pushing anonymous data they will simply stop.

This file was before on a gist - I suggested to move it under quarkus.io so its all as transparent as possible.

assets/json/03656937-19FD-4C83-9066-C76631D445EA.json
maxandersen: also, wouldn't it make more sense to have this under `quarkus.io/uages/xx-yyy.json` ?

brunobat: this is remote config for us... We don't instruct users to change that.

gsmet:
@brunobat and it won't break us in case of a GitHub outage, right?

brunobat:
The builds themselves will check each 12h, if the file is not found will inactivate analytics and will try later.

maxandersen:
should fail? you mean should NOT fail, right?

brunobat:
It will actually fail because the remote config file is kind of important. 
I can do make it a manual test...

maxandersen:
We should fix that asap. Just because quarkus can't get info about its analytics should not prevent users from building their project.

brunobat:
User's project will continue to build.
It's the Quarkus project itself that has a test that will fail if the config is not there.

PR #1734, opened 2023-06-27
Improve usage policy a bit

gsmet:
@maxandersen could you have a look?

### DIFF
--- a/_includes/usagepolicy.html
+++ b/_includes/usagepolicy.html
@@ -4,7 +4,7 @@
       <p>This page contains the policy for the Red Hat sponsored anonymous data collection. See <a href="/usage">Usage</a> page for details.</a></p>
       <p class="intropara">In connection with your use of this Red Hat offering, Red Hat may collect telemetry data about your use of the software. This data allows Red Hat and community extension writers to understand how the software is being used and to improve Red Hat offerings and support, including identifying, troubleshooting, and responding to issues that impact users. The data may be shared internally within Red Hat to improve the user experience.</p>
       <H3>What information does Red Hat collect?</H3>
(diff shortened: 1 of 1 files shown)


_includes/usagepolicy.html
```diff
@@ -4,23 +4,23 @@
       <p>This page contains the policy for the Red Hat sponsored anonymous data collection. See <a href="/usage">Usage</a> page for details.</a></p>
       <p class="intropara">In connection with your use of this Red Hat offering, Red Hat may collect telemetry data about your use of the software. This data allows Red Hat and community extension writers to understand how the software is being used and to improve Red Hat offerings and support, including identifying, troubleshooting, and responding to issues that impact users. The data may be shared internally within Red Hat to improve the user experience.</p>
       <H3>What information does Red Hat collect?</H3>
-      <p>Tools within the software monitor various metrics and this information is transmitted to Red Hat.  Metrics include the following information such as:</p>
+      <p>Tools within the software monitor various metrics and this information is transmitted to Red Hat. Metrics include the following information:</p>
```
gsmet, line 7: This is not syntactically correct so it needs a fix, be it this one or something else.

_includes/usagepolicy.html
```diff
@@ -4,23 +4,23 @@
...
       <ul>
         <li>Usage information about components, features and extensions.</li>
         <li>The number and nature of errors that occur during development.</li>
         <li>Quarkus version</li>
         <li>Java version, vendor</li>
         <li>Build tool used</li>
         <li>Build duration</li>
-        <li>List of extensions and their versions</li>
-        <li>Timezone</li>
+        <li>List of <b>community</b> extensions and their versions</li>
```
maxandersen, line 15: since we also report productized extensions would something like this be better:

```suggestion
        <li>List of <b>public</b> extensions and their versions</li>
```

PR #1743, opened 2023-07-06
List Jakarta, MicroProfile and OpenTelemetry specifications supported by Quarkus

radcortez:
The idea here is to list all specifications supported by Quarkus, because we are not fully compliant with all platforms and specifications, and sometimes users are not aware which things we support or not. Ideally, this should be auto-generated, but for now, is manual to put the idea forward to see if we should pursue this.

Also, there are other specifications that we support missing, but we can add them later.


radcortez:
Here is an example:

Screenshot 2023-07-07 at 00 19 13

radcortez:
Due to the natural lag between a spec being out and being adopted, it may confuse the user, thinking that we support the most recent version of something.

We have also other situations, where we deliberately choose not to support a spec beyond a certain version (MP Metrics for instance)

hamburml:
> I wonder if this should be per version?

Maybe per LTS Version?

hamburml:
This helps me immensely when writing libraries. I want to add jakarta ee or microprofile apis instead of quarkus dependencies and the library is easier to maintain. For example, we currently have the problem that our library needs to be updated if there are any CVEs - if we develop against the spec, only the service itself needs to be updated.
With such a page, it is easier to choose which version is supported by the Quarkus version.

> We have also other situations, where we deliberately choose not to support a spec beyond a certain version (MP Metrics for instance)

Could you document this also?

radcortez:
> Does it make sense to add the patch level of the specs?

I thought about it, but the idea of this page was to just be a high-level view of the supported specifications.

> Could you document this also?

This was announced here: https://quarkus.io/blog/quarkus-observability-roadmap-2023/

I'm not sure if this is the best location for it. Again, the idea was to just be a high-level view. While I understand that this is relevant information, then we would also need to document why we don't support the unlisted Jakarta specifications.

hamburml:
> Again, the idea was to just be a high-level view. 

Sad, I would love a detailled overview of the most important specs per LTS version.

>  While I understand that this is relevant information, then we would also need to document why we don't support the unlisted Jakarta specifications.

Which is a good thing because it is transparent and open I would say.

In my eyes, however, this is already the right step, which can of course be expanded in subsequent steps.

radcortez:
I think we can add that information somewhere, but not here, because this is intended to be a summary page.

Maybe on a WIKI page?

maxandersen:
+1 for overview page. 

Could the detailed info be a page generated in the docs based on what dependencies we use so no need to maintain it ?

radcortez:
> Could the detailed info be a page generated in the docs based on what dependencies we use so no need to maintain it ?

Yes, it should be possible.

gmuth:
I appretiate the documentation effort. Still, I wonder why quarkus is not fully compliant with the microprofile specifications.

radcortez:
> I appretiate the documentation effort. Still, I wonder why quarkus is not fully compliant with the microprofile specifications.

Quarkus is fully compliant with MP 4.x and 5.x. With MP 6.x, we made the hard decision not to support MP Metrics, but everything else is compliant (including previous versions of MP Metrics). We announce our plans around Observability, including the reason for us not supporting the new MP Metrics coming with MP 6.x: https://quarkus.io/blog/quarkus-observability-roadmap-2023/

Starting with MP 7.x, the platform will drop MP Metrics in favor of OTel Metrics, allowing us to claim full compatibility again. You can follow the progress, TCK results, and Compatible Certification Requests here: https://github.com/quarkiverse/quarkus-microprofile

PR #1747, opened 2023-07-09
Drop Gemfile.lock as it requires exact bundler version

michalvavrik:
I couldn't figure why I can't serve site locally till I tried Docker-way that also fails - when you use different Bundler version, they can be incompatible with `Gemfile.lock`. IMO everyone should build his own lock file following README or there must be one common Bundler version used by everyone. I checked CI, but I'm not sure if it won't break anything.

```bash
[mvavrik@apc-ap8641-3a-d11d quarkusio.github.io]$  docker-compose up
Starting quarkusiogithubio_jekyll_1 ... done
Attaching to quarkusiogithubio_jekyll_1
jekyll_1  | Bundler 2.4.14 is running, but your lockfile was generated with 2.2.16. Installing Bundler 2.2.16 and restarting using that version.
jekyll_1  | There was an error installing the locked bundler version (2.2.16), rerun with the `--verbose` flag for more details. Going on using bundler 2.4.14.
jekyll_1  | Could not reach host index.rubygems.org. Check your network connection and try
jekyll_1  | again.
quarkusiogithubio_jekyll_1 exited with code 17
```


michalvavrik:
Ouch, this won't do.

maxandersen:
The lock file is *intended* to committed so we have reproducible way of building. 

The docs/readme explain how to use separate env tools to have isolated build.

michalvavrik:
@maxandersen I read README more than once and I'm only able to build site when I delete lockfile. And I tried it several times on new forks. It goes down to exact version that is not documented. Plus Docker way didn't work for me at all. Adding `BUNDLED WITH` to lockfile will cause troubles unless there is documented version of bundler. Only thing in README that is remotely specific to version is `If you use rvm, run: rvm use 2.7.1` to which I answer - I don't use rvm on Fedora.

> The docs/readme explain how to use separate env tools to have isolated build.

Please point me to it.

maxandersen:
Rvm is a way to do isolated builds. 

If you don't want to use that I'm not sure how you build ruby projects on fedora without something similar. 

I'll test the dockerfile as it should work but been a while since I tried.

michalvavrik:
Now I understand, for me, it wasn't clear 2.7.1 is way to go and there is possibility I could run into a trouble with other versions (there is `If you use rvm`, I'd make it recommendation instead). It is fine if noone else read it as I do. I have no issues with building it locally now.

PR #1780, opened 2023-09-14
About content update

insectengine:
Adding content suggested by a member of the community (@chrischiedo) to the **What is Quarkus** page.

This is marked as a draft so the team can review the content and wordsmith it as necessary.


chrischiedo:
Thanks @insectengine for this.

I think we can reformat the added content so that it has a subtitle that aligns with the rest of the page (rather than having the heading on the right side of the icon):
Screenshot 2023-09-15 at 11 44 37

(I actually pulled down your PR and made the change locally, but it seems like I don't have permission to push back up to your PR).

insectengine:
Moved title as @chrischiedo  suggested. 

@maxandersen  @n1hility  - any thoughts on the content?

marcelstoer:
@insectengine May I ask why you closed this? IMO it was helpful information.

insectengine:
Reopened. Was cleaning up old branches and removed inadvertently.

maxandersen:
missed this - yeah, the draft label made me miss it :)

This aspect of building on existing libraries is generally what we refer to as "Best of Breed Libraries" -  I would suggest using that rather than "Community" for this section.

About the actual content - we can pick on many different libraries, so hard to choose, but the set here does seem to be a pretty good starting set. Apache Camel stands out a bit though as its not core to quarkus but its for sure a big part of the full platform.

@cescoffier thoughts?

PR #1783, opened 2023-09-15
Add subtitle to reactive/continuum page

chrischiedo:
Draft PR to add subtitle to the reactive/continuum page. Unlike the other sections (developer joy, container first, standards, etc), the reactive section was missing a subtitle (not sure whether that was by design or just an oversight?). Also, updated the second paragraph with punctuation + wording corrections.

cc: @insectengine


insectengine:
@chrischiedo  - Looks good to me.

@gsmet  - thoughts?

PR #1810, opened 2023-10-09
Add BR to the Portuguese menu items

gastaldi:
- This makes pt.quarkus.io explicit about being pt_BR and not pt_PT

### DIFF
--- a/_data/projectfooter.yaml
+++ b/_data/projectfooter.yaml
@@ -66,7 +66,7 @@ links:
         code: en
         url: https://quarkus.io/
       - page: Português (Brasileiro)
-        code: pt
+        code: pt-br
         url: https://pt.quarkus.io/
       - page: Español
         code: es
--- a/_data/projectfooter.yaml
+++ b/_data/projectfooter.yaml
@@ -65,7 +65,7 @@ links:
       - page: English
         code: en
         url: https://quarkus.io/
-      - page: Português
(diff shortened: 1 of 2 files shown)


holly-cummins (approved):

gastaldi:
Should we keep the longer `Brasileiro` name or just `BR`? Either works for me but I think the latter is better since it takes less space, WDYT?

_data/projectfooter.yaml
```diff
@@ -65,8 +65,8 @@ links:
       - page: English
         code: en
         url: https://quarkus.io/
-      - page: Português
-        code: pt
+      - page: Português (Brasileiro)
+        code: pt-br
```
maxandersen, line 69: im surprised the footer still link to pt.quarkus.io after this change :)

maxandersen:
> Should we keep the longer `Brasileiro` name or just `BR`? Either works for me but I think the latter is better since it takes less space, WDYT?

(BR) fits better. +1 on keeping it short.

_data/projectfooter.yaml
gastaldi: The url is set below this line, I believe this is used in the hreflang metadata only

maxandersen: ah of course. gotcha.

brunobat (approved):

Issue #1352, opened 2022-03-03
Collapsible table of content for guides


metacosm:
Not really a bug, more like a feature request. It would be nice to have the ToC always available regardless of the spot in the page as it is annoying to have to scroll all the way back up when exploring sections of long guides. Ideally, the ToC would be collapsible so that it could be hidden when not needed.

metacosm:
Similarly, it could be interesting to have the top menu always available as well.

gsmet:
/cc @insectengine 
We should probably move the TOC to a menu button outside of the page content itself when scrolling down.

metacosm:
Ideally, the ToC should be accessible from anywhere in the page just by clicking on a collapsed element on the side of the page (so only motion required would be to move the mouse to the edge of the window and click, without having to move up or down and/or have to hit a small target).

insectengine:
I'll put some thought into this and try to sort out a workable solution.

Sticky navigation can be problematic if the contents get too long and render past the viewport. To defeat this, it would require the authors to organize the content into appropriate sections for some sort of accordion/collapsable structure.

One of the advantages of using a sticky TOC would be a way finding element so the user knows where they are at any given time. This is negated if the TOC is hidden (as suggested being out of they way via a button). Using such a button would mean that the TOC would cover readable content when being used. Then this brings up another issue is what's visible by default when a user hits the page? Is the TOC open and covers the content or is it closed and the user may miss it being there and doesn't see the full scope of the doc.

I believe this is why many choose to do a side bar navigation like IBM uses: https://cloud.ibm.com/docs/cloud-infrastructure

metacosm:
Whatever works as long as the navigation is constantly available without having to scroll up and down to move between sections.

insectengine:
Thanks for surfacing this. I'll put it on my Q4 goals for completion.

insectengine:
https://github.com/quarkusio/quarkusio.github.io/pull/1814

PR #1814, opened 2023-10-12
refactor TOC to be sticky

insectengine:
Addresses: https://github.com/quarkusio/quarkusio.github.io/issues/1352

1. Re arranged page to have TOC on the left
2. created two column layout (removed floating TOC) so scrolling content wouldn't clash with TOC. This also gives a more comfortable line length for reading than traditional full width content.
3. It's sticky so it stays on the top of the page when the page scrolls
4. There is a max height on the TOC with Scrolling. This allows the TOC to scroll when there is more links that the TOC view window (only happens on one guide I've found so far)


metacosm:
A step in the right direction but why is there so much white space on the left side of the TOC? This seems like wasted space to me… that space exists even when the window is fairly narrow.

insectengine:
Are you talking about this space?
thisspace

insectengine:
@metacosm  The website isn't full width, it's centered with margins on either side that vary in width as the site is views on smaller screens. It starts at 13rem and has a medium size for tablets and small for mobile. We do this to prevent the site from exploding into gibberish on a huge screen. Switching from this to a full width for only the guide pages would be extremely jarring for the user.
border2
border

insectengine:
@metacosm  - Changing the breakpoint for the spacing would be site wide. @ebullient  also suggested we fine tune some of the font sizes and whatnot at the mid range. We can do that but I think that should be a separate exercise and PR.

insectengine:
Erin suggested moving the TOC to the right side of the page (where it originally was) so the important content is on the left (as read by left-right languages). I'm making this change and looking into adding javascript to track the users place in the TOC.

metacosm:
Yes, the TOC on the left somehow bothered me but I couldn't decide if it was just because I wasn't used to it yet or something so I decided to let it slide 😉

insectengine:
I'm going to not try and put a scroll spy indicator on the TOC as it's beyond my depth. If we have someone on the team skilled in JS that could make it work, that would be great.

ebullient (approved):

metacosm (approved):
Thanks!

GabrielFoxDev:
Just opened an old tab to have it refreshed and greet me with that nice sticky ToC! :smile: 
Thanks!

PR #1826, opened 2023-10-30
Exclude all versioned guides (only latest guides are left)

ynojima:
Version is added time to time. Not to break the preview site deploy by exceeding surge.sh size limit, exclude all the versioned guides except latest guides.

(Japanese localized site preview is affected by recent minor release: https://github.com/quarkusio/ja.quarkus.io/actions/runs/6686496521/job/18165982368?pr=1087)


gastaldi (approved):
Good idea

PR #1830, opened 2023-11-08
Fix author for 2023-11-08-quarkus-3-2-8-final-released.adoc

rsvoboda:
Defined author is alexeyloubyansky
See https://github.com/quarkusio/quarkusio.github.io/blob/develop/_data/authors.yaml#L92

Current state (no author details) for 3.2.8 announce compared to 3.5.0 one:

Screenshot 2023-11-08 at 20 54 41


aloubyansky (approved):
Thanks @rsvoboda !

PR #1837, opened 2023-11-17
Update cheatsheet filter

marko-bekhta:
It looks like the filter value changed on the https://developers.redhat.com side, and the current link gives 0 results...


gsmet (approved):
Thanks!

PR #1846, opened 2023-11-29
add governance page and links to threads and socials page

insectengine:
Added a governance page using the Narayana project's as a base. The content needs to be reviewed and updated to better reflect the Quarkus project.

### DIFF
--- a/_data/projectfooter.yaml
+++ b/_data/projectfooter.yaml
@@ -22,10 +22,16 @@ links:
         url: /security
       - page: Usage
         url: /usage
+      - page: Governance
+        url: /governance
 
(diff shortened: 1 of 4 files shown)


insectengine:
@paulrobinson @maxandersen @n1hility

maxandersen:
I agree we should have a page like this but this is showing a governance model we shouldn't be having - its way too Red Hat focused.

insectengine:
@maxandersen  - agreed. this content is a starting point. Can you come up with acceptable copy? Or get someone with knowledge to volunteer?

insectengine:
Out of date.

PR #1849, opened 2023-11-30
Fix incorrect quarkus.analytics.disabled boolean value

### LINKED ISSUE #1736: Wrong boolean value in Quarkus Analytics property example
**Describe the bug**

I think the last sentence on the `Quarkus Usage Analytics` page is wrong:
https://github.com/quarkusio/quarkusio.github.io/blob/fe46373b21251161a3345be2f5de2210d3d936d7/usage.md?plain=1#L45-L53

If I understand it correctly, we need to set the value do `false` to enable it, so it would be:
```
Similarly, you can set it to `quarkus.analytics.disabled=false` if you want to enable it for a single run.
```

**Which pages**

[Quarkus Usage Analytics](https://quarkus.io/usage/)

gefloh:
The boolean value for `quarkus.analytics.disabled` is incorrect here. Fixes #1736 .

### DIFF
--- a/usage.md
+++ b/usage.md
@@ -50,4 +50,4 @@ Example using maven:
 mvn clean install -Dquarkus.analytics.disabled=true
 ```
 
-Similarly, you can set it to `quarkus.analytics.disabled=true` if you want to enable it for a single run.
+Similarly, you can set it to `quarkus.analytics.disabled=false` if you want to enable it for a single run.


gefloh:
@gsmet can you have a quick look? This doesn't take long. Thanks.

gsmet (approved):
Right. Your change looks good.

I think this is a bit confusing though and given the importance of it, we should clarify things. I'll have a closer look.

gsmet:
Thanks!

PR #1856, opened 2023-12-11
Virtual threads - blog post 6

cescoffier:


maxandersen:
...i'm reading this and wondering what is specific in this to virtual threads? this is how you containerize (almost) any java/quarkus app or am I missing something?

cescoffier:
No, you are not missing anything. We made it plain boring. I might have to highlight that.

maxandersen:
> No, you are not missing anything. We made it plain boring. I might have to highlight that.

yeah - if that is the intention I would definitely highlifht it. 

Possibly even consider not making that separate post and just point to info how to deploy in docs and focus on the different aspect going forward?

cescoffier:
I got asked for that specific post. So I believe there is a need.

cescoffier:
Maybe I should copy that post to the docs (as ... it's not crystal clear (too many possibilities))

PR #1862, opened 2023-12-19
Update date suffix in blog post about 3.7, 3.8 and 3.9 releases

gemmellr:
31th - > 31st


gsmet (approved):
Ah, yes, right, thanks!

PR #1872, opened 2024-01-15
Fix search returning no results after explicitly selecting "All categories"

yrodiere:
See reproducer in this video:

[Screencast from 2024-01-15 11-48-38.webm](https://github.com/quarkusio/quarkusio.github.io/assets/412878/358b79a9-4b54-4bfd-81dc-2847ac871e5d)

### DIFF
--- a/assets/javascript/guides-app.js
+++ b/assets/javascript/guides-app.js
@@ -69,12 +69,18 @@ const app = createApp({
     },
     category: {
       get() {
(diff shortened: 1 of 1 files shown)


yrodiere:
cc @ia3andy , in case your reimplementation has a similar problem (I doubt it, but what do I know)

yrodiere:
Works on the preview; I'll merge.

PR #1895, opened 2024-02-21
Use qs web component

ia3andy:
This is a draft until I come back from PTOs

It mostly need CSS cleaning to be ready..

The search service needs to be started locally, and it might require some change in the Chrome config to allow cors on localhost.


ia3andy:
This needs: https://github.com/quarkusio/search.quarkus.io/pull/58

_config.yml
```diff
@@ -27,7 +27,9 @@ github_fork_url: "https://github.com/quarkusio/quarkus"
 language: en
 search:
   # The URL of the remote search service
-  url: "https://search.quarkus.io/"
+  script: "/assets/javascript/search-wc.js"
+  script-copy: "https://search.quarkus.io/static/bundle/main.js"
```
yrodiere, line 31: Shoudn't this rather be `script-to-copy`? It's not itself the copy... ?

_config.yml
```diff
@@ -27,7 +27,9 @@ github_fork_url: "https://github.com/quarkusio/quarkus"
 language: en
 search:
   # The URL of the remote search service
```
yrodiere, line 29: This comment is outdated :]

_includes/index-doc-item.html
```diff
@@ -1,17 +1,21 @@
...
-</div>
+<qs-guide type="{{ include.type | escape }}"
+          title="{{ include.title | escape }}"
+          url="{{ guide_url }}"
+          summary="{{ include.summary | markdownify }}"
+          keywords="{{ include.keywords | escape }}"
+          categories="{{ include.categories | escape }}"
+          origin="{{ include.origin | escape }}"
+>
+    <!-- WE KEEP PART OF THE CONTENT FOR SEO -->
```
yrodiere, line 17: Why not keep the whole content while we're at it? It would be easier for debugging quarkus.io when one doesn't have an instance of search.quarkus.io running locally.

_includes/index-docs.html
```diff
@@ -3,162 +3,149 @@
...
-      <div class="grid-wrapper">
-        <div class="grid__item docs-card pdfbkg">
-          <h4><a href="https://lordofthejars.github.io/quarkus-cheat-sheet/" target="_blank">Quarkus Cheat Sheet</a></h4>
-          <div class="description">Download full cheatsheet as PDF. <a href="https://developers.redhat.com/search?t=quarkus&f=type%7Echeat_sheet" target="_blank">
-            Get more cheatsheets on the Red Hat Developers website. <i class="fas fa-external-link-alt"></i></a></div>
-          <div class="keywords"></div>
-          <div class="categories"></div>
+
+    <!-- Static content displayed when there is no input in the search form or Javascript is disabled,
+         but also used as a source of data for cards displayed by the vue.js app. -->
```
yrodiere, line 53: Looks like your reformatted this file from 2-space indent to 4-space indent, and it's making the diff rather hard to read :(

_plugins/copy-search-wc.rb
```diff
@@ -0,0 +1,37 @@
...
+            content = response.read
+            if response.status[0] != '200'
+              raise "Error: The search wc '#{script_copy_url}' is not accessible. Status: #{response.status[0]} - #{response.status[1]}"
+            elsif content.nil? || content.empty?
+              raise "Error: The search wc '#{script_copy_url}' is empty."
+            else
+              content_without_sourcemaps = content.gsub(/\/\/# sourceMappingURL=.*\.map/, '')
+              File.delete(assets_path) if File.exist?(assets_path)
+              File.open(assets_path, 'w') { |file| file.write(content_without_sourcemaps) }
+              puts "The search wc '#{script_copy_url}' has been copied to '#{assets_path}'."
```
yrodiere, line 28: Ok I'll bite... what does "wc" stands for? I sure hope it's not what I have in mind :)

yrodiere (commented):
Thanks. I added a few comments below.

_plugins/copy-search-wc.rb
ia3andy: lol 🚽

more seriously web-component

_includes/index-doc-item.html
ia3andy: I am not sure I understand the plus of adding the whole content. It would make the page longer to load, also it's not visible anyway.

yrodiere: > It would be easier for debugging quarkus.io when one doesn't have an instance of search.quarkus.io running locally.

But okay, fair point about the page being longer to load.

_config.yml
ia3andy: I've changed to `script-mode=cached/direct`

ia3andy:
Ok it works and can be tested in the preview 🥳
cc @marko-bekhta @yrodiere

yrodiere:
Thanks Andy.

I noticed a few things:

1. The ellipses ("...") seem to be gone in the excerpts from the content:
image
image
2. Selecting version "3.8" or earlier leads to a 404... but maybe that's just a limitation of the preview? Regardless, I wasn't able to test the "old" guide list.
3. If you go to https://quarkus-site-pr-1895-preview.surge.sh/guides/ , then select version "Main - Snapshot", then click "back" to go to the previous page in your browser... the list displays a search, for some reason, instead of just the default display by guide type (tutorial/etc.). The value of the drop-down also doesn't get reverted, but *that* seems to be a pre-existing bug. The first problem isn't, though.
image
4. Calls to the search backend appear to be twice as numerous as before: first an OPTIONS call, then a GET call. And that's not just the first call either, but every single call. Is that on purpose? The existing version only does a GET call.
Note that we sometimes have serious latency problems on this cluster, so doubling the number of calls could have dramatic impact...
image

marko-bekhta:
One more thing 🙈 if a user tries to search for something and then deletes the search text a never-ending-spinner is displayed:
image

ia3andy:
@yrodiere https://stackoverflow.com/questions/29954037/why-is-an-options-request-sent-and-can-i-disable-it

ia3andy:
@marko-bekhta good catch, I never experienced it because there is no latency locally :)

yrodiere:
> @yrodiere https://stackoverflow.com/questions/29954037/why-is-an-options-request-sent-and-can-i-disable-it

Thanks, but we didn't experience that before, so surely there's been a change that triggers it. Is that change worth doubling the latency?

ia3andy:
For the `...`, I think it should be done by the api instead if that's possible?

ia3andy:
> > The ellipses ("...") seem to be gone in the excerpts from the content:
> 
> Will be fixed in [quarkusio/search.quarkus.io#223](https://github.com/quarkusio/search.quarkus.io/pull/223) , thanks @marko-bekhta .

🙏

> 
> > If you go to https://quarkus-site-pr-1895-preview.surge.sh/guides/ , then select version "Main - Snapshot", then click "back" to go to the previous page in your browser... the list displays a search, for some reason, instead of just the default display by guide type (tutorial/etc.). The value of the drop-down also doesn't get reverted, but _that_ seems to be a pre-existing bug. The first problem isn't, though.
> 
> Still the same problem: [Screencast from 2024-04-08 15-23-22.webm](https://github.com/quarkusio/quarkusio.github.io/assets/412878/3e15f2cb-51ae-4048-9c75-692f56655fb9)
> 

I can't reproduce this issue because I suspect it's a latency issue. I have a good idea of why: the search form detects the change which starts a search, and then loads the page. I have a fix and I am creating a PR for it on search.quarkus.io

> > Selecting version "3.8" or earlier leads to a 404... but maybe that's just a limitation of the preview? Regardless, I wasn't able to test the "old" guide list.

Yes I tested it locally and it works fine

> 
> Still the same problem -- is this just a limitation of the preview? Did you test "old" guide lists locally, e.g. the one for 2.7?
> 
> > Calls to the search backend appear to be twice as numerous as before: first an OPTIONS call, then a GET call
> 
> Fixed, thanks.

yeahh!

ia3andy:
@yrodiere so when the latest PR is in prod, we can merge this right?

yrodiere:
> > > If you go to https://quarkus-site-pr-1895-preview.surge.sh/guides/ , then select version "Main - Snapshot", then click "back" to go to the previous page in your browser... the list displays a search, for some reason, instead of just the default display by guide type (tutorial/etc.). The value of the drop-down also doesn't get reverted, but _that_ seems to be a pre-existing bug. The first problem isn't, though.
> > 
> > 
> > Still the same problem: [Screencast from 2024-04-08 15-23-22.webm](https://github.com/quarkusio/quarkusio.github.io/assets/412878/3e15f2cb-51ae-4048-9c75-692f56655fb9)
> 
> I can't reproduce this issue because I suspect it's a latency issue. I have a good idea of why: the search form detects the change which starts a search, and then loads the page. I have a fix and I created a [PR](https://github.com/quarkusio/search.quarkus.io/pull/228) for it on search.quarkus.io

Seems to work now, thanks

yrodiere (approved):

yrodiere:
LGTM thanks; feel free to merge whenever you have time to handle the possible consequences :)

ia3andy:
🤞

PR #1896, opened 2024-02-26
Update quarkiverse guide links

marko-bekhta:
I also noticed that Dynamo and S3 guides were duplicated in data and cloud categories so I've combined those... 
Let me know if this makes sense


yrodiere:
Makes sense to me, thanks. I'll let @gsmet merge though, as he generally notices a problem I completely overlooked just after I merge things like this ;D

gsmet (approved):

gsmet:
Yeah, the links for 2.13 are probably not ideal but we don't have any history on the Quarkiverse so we will have to live with it.

Thanks!

marko-bekhta:
😭 😔 
image

Now AWS guides are back to using `dev` instead of `main`

PR #1901, opened 2024-02-27
fix imagesdir in the Kiota post

andreaTP:
The images are not showing, I believe this is the fix:
https://quarkus.io/blog/quarkus-kiota/


maxandersen (approved):

andreaTP:
cc. @maxandersen

PR #1910, opened 2024-03-11
add author page linked from blog

maxandersen:
as suggested by @Sanne I added a page per authors as he was looking for @FroMage 's posts.

this uses https://github.com/gouravkhunger/jekyll-auto-authors which is explicitly solving this issue
as its missing feature of built in archive/pagination.

~~we have `/blogs/tag/<tag>` so added `/blogs/author/<author>` with same layout
and added bio/twitter/headshot to the page.~~

added `/author/<author>` with same layout as tags but lets us have quarkus.io/author landing/index page


gsmet:
@insectengine I let you have a look at the layout?

PR #1943, opened 2024-04-15
Disable the in-browser autocomplete for the search input

marko-bekhta:
Fixes https://github.com/quarkusio/search.quarkus.io/issues/231

I think we should keep the `q` name, as it's also the API parameter. And if we'd wanted to have autocomplete working for the field then we should consider something backed by the search service.. though I'm not sure it would be much helpful in this case.

### DIFF
--- a/_includes/index-docs.html
+++ b/_includes/index-docs.html
@@ -38,7 +38,7 @@
           </select>
         </div>
         <div class="search">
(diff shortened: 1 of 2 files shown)


yrodiere:
Rebased and merged, thanks

PR #1944, opened 2024-04-16
Fix fallback dev search

ia3andy:
This in combination with https://github.com/quarkusio/search.quarkus.io/pull/238 should fix the local dev search issue.

The idea is to use the production search script and host by default in dev. Using it will result in a 403 leading to using the local search fallback.

### DIFF
--- a/_config_dev.yml
+++ b/_config_dev.yml
@@ -3,10 +3,10 @@
 # We don't want to use the production instance of search.quarkus.io for development.
 search:
 
-# By default we use the staging cluster in cached mode, you can change it to use your local instance if running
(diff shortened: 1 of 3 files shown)


ia3andy:
cc @gsmet @yrodiere @marko-bekhta

yrodiere:
@ia3andy Can't you just use staging, since prod will always fail due to CORS?

See https://github.com/quarkusio/search.quarkus.io?tab=readme-ov-file#current-process

ia3andy:
@yrodiere I also get CORS on staging

yrodiere (approved):

yrodiere:
merged, thanks

Issue #1950, opened 2024-04-18
Illegible colour scheme in the code snippets


meridsa:
# Description

The code snippets are challenging to read. I am unsure how much I am uncommonly affected due to red-green colourblindness, but I feel that both comments (gray text on gray background), as well as the annotation and property (orange-on-gray) are difficult to read. I have also checked with the [silktide plugin](https://chromewebstore.google.com/detail/silktide-accessibility-ch/mpobacholfblmnpnfbiomjkecoojakah) for these colours

# Show issue

## annotation text (i.e. "/service" text)
![image](https://github.com/quarkusio/quarkusio.github.io/assets/35993351/851bf373-1681-4279-8e10-4cff913dd89d)

## Comments
![image](https://github.com/quarkusio/quarkusio.github.io/assets/35993351/9ba084af-fc48-4af1-a68a-b65944bc20ff)

# Scope

Affects all code snippets on all pages.

gsmet:
This is probably a bug in the dark mode introduced yesterday. @insectengine could you have a look soon? It seems very problematic.

meridsa:
I changed to dark-mode to be able to read:

![image](https://github.com/quarkusio/quarkusio.github.io/assets/35993351/ad5a35ba-c190-41b5-a6a5-0483fe8ac927)

PR #1908, opened 2024-03-05
Darkmode

insectengine:
Added a clean dark mode for the site. To keep it simple, it detects the OS preference and sets the site via CSS.

Issue #1809, opened 2023-10-03
Enhance Quarkus Docs page with Dark mode


gian1200:
### Description

As stated in the title, to enable a Dark mode option for the docs page. Nothing fancy nor complicated, but focusing on readability.

OS sync for initial configuration would be a nice to have thing, but not necessary.

Current docs:
https://quarkus.io/guides/cache
![image](https://github.com/quarkusio/quarkus/assets/405347/9c959ebf-2ac4-476b-a0ae-f1caa171b5db)

### Implementation ideas

Maybe something similar to GitHub's Dark dimmed: grey (non black background and non white text).

GitHub:
![image](https://github.com/quarkusio/quarkus/assets/405347/9b5641c7-a806-41d0-9b8b-74d9580f8b50)

I know colors are a preference and relative, but if some people like a darker (or lighter) mode, maybe it can be expanded to:

- Lighter (whiter or current)
- Light (current or darker current)
- Dark (grey)
- Darker (black)

gsmet:
I moved the issue to the website project. /cc @insectengine this might be a good idea. If I'm not mistaken @jponge had fun with this for the Mutiny website: https://smallrye.io/smallrye-mutiny/2.5.1/ .

jponge:
Mutiny uses [MkDocs Material](https://squidfunk.github.io/mkdocs-material/) so the dark mode was built-in and I _just_ had to choose colors 🤣

holly-cummins:
Watching this for http://extensions.quarkus.io and http://docs.quarkiverse.io, too.

insectengine:
Put this on my work queue to look at but will need to prioritize.

This would need to be a site wide thing (not just on docs) so it will be a larger lift.

roby2014:
Bump, need this.

The-Funk:
I get the occasional light induced migraines so dark mode is enabled on my machine everywhere I can find it. Glad to see someone already put in this request.

insectengine:
I submitted a draft PR with a dark mode for the site. It's a clean approach that detects the user's OS settings and sets either light or dark mode. I didn't include a switcher to avoid extra javascript and visual clutter on the page. 

https://github.com/quarkusio/quarkusio.github.io/pull/1908

Note" While I've created dark mode versions of images on the content pages, we'll have to sort out how to swap images in the documentation to have dark mode friendly images via ascidoc.

@gsmet @maxandersen @holly-cummins

maxandersen:
[...]

PR #1954, opened 2024-04-18
Reinstate dark mode framework (with dark mode disabled)

holly-cummins:
https://github.com/quarkusio/quarkusio.github.io/pull/1951 reverted #1908, to resolve https://github.com/quarkusio/quarkusio.github.io/issues/1950 (which affected light mode), and also to solve some issues with dark mode.

- search icons are not showing
- in some guides, images with a transparent background do not render well

However, it's desirable to have the #1908 changes in the code base. One reason is that we want to fix these minor issues and have dark mode, because [people ask for it](https://github.com/quarkusio/quarkusio.github.io/issues/1809#issuecomment-1773616057). Doing these fixes will be easier, and lower risk, if there's not a major merge to bring back #1908. Another reason is that #1908 removes some hardcoded colours and introduces more semantically-meaningful variable names, which is a good thing even if we never do dark mode. 

This PR reintroduces #1908, but with the code background colour corrected, and with dark mode always-off. 

image

We should do a thorough visual review of the preview to make sure there are no other light mode regressions (doing this kind of colour-to-variable refactoring can be error-prone, which is why we want to only do it once!)

cc @insectengine


holly-cummins:
Next steps after this merges: 
- Add a toggle for dark mode (possibly hidden, at first)
- Fix search icon rendering in dark mode 
- Adjust 'extensible' content such as guides and blogs to put in a light-coloured background on all images. This ensures all images are legible (we could eventually also introduce a naming convention to allow an override for images which are declared to work in both modes and want a transparent background)
- Carefully review dark mode for other visual problems

maxandersen (approved):
+1 code blocks is readable and dark mode is disabled so can roll out more safely

holly-cummins:
Oh, and the other next step is to replicate the toggle mode to http://quarkus.io/extensions, which is at the moment using system settings, and should be switched to light mode (while waiting for the toggle)

maxandersen:
examples of pages having "bad" darkmode:

http://0.0.0.0:4000/blog/quarkus-kiota/ - code block with darkgrey on grey



http://0.0.0.0:4000/blog/quarkus-saas-backend-time-tracking-app-case-study/ - transparent logo unreadable

gsmet:
@maxandersen your comment above is not about this PR, right, it's when you enabled dev mode yourself? We can merge it?

holly-cummins:
> @maxandersen your comment above is not about this PR, right, it's when you enabled dev mode yourself? We can merge it?

Answering for @maxandersen, we don't have a good place to record the task list for dark mode, so he's put it on this work item for now. All of the problems he mentioned are technically in this changeset, but could never be seen by a user. But I will let @maxandersen also answer for himself!

maxandersen:
Correct - it was just when i enabled dark mode. I added it here as way to collect the info.

For the actual PR I'm all +1 - code reads, looks and behaves fine.

PR #1972, opened 2024-05-07
Fix compose.sh on Linux

yrodiere:
@maxandersen I don't get why the variable was commented out, probably a mistake?

As things are, your fix just makes it work for Mac, but no longer on Linux.

With this patch it works on Linux again, but I'll let someone else check it works on Mac...


maxandersen:
lets just revert it - using latest podman version that honors/ignores :z is better anyway.

opened https://github.com/quarkusio/quarkusio.github.io/pull/1972

PR #1975, opened 2024-05-12
Increase RSS feed to 50 last entries

emmanuelbernard:
Also update property name to latest from jekyll-feed

Receiving only the last 10 is detremental for people
only occasionally fetching RSS feeds through their reader.
We could go infinite but 50 bring up back 6 months in the past
that is a good start.


emmanuelbernard:
I tested locally and RSS feed now has more entries 

```
http http://127.0.0.1:4000/feed.xml | grep title
```

emmanuelbernard:
Force push to sign commit

gsmet (approved):

gsmet:
Makes sense, thanks!

PR #1983, opened 2024-05-17
add light/dark mode to homepage

### LINKED ISSUE #1980: Should we have some light-mode content on the front page?
If I'm on http://quarkus.io and I change the dark mode toggle, nothing on the page changes. That's consistent with the design, but it makes it look like the toggle isn't working. I wonder if something on the page should be light mode, so that users get visual feedback on the toggle? (I got confused and thought I'd broken something because I had a sun image and dark content, but it was just that the whole page is always dark.)

cc @insectengine

insectengine:
Made the necessary changes to enable dark/light mode on the home page

resolves #1980


holly-cummins:
Sweet!

gsmet (approved):
I force pushed to fix the conflict.

PR #1878, opened 2024-01-19
Add edit-this-page link

marcelstoer:
This started at https://quarkusio.zulipchat.com/#narrow/stream/187030-users/topic/.E2.9C.94.20How.20does.20Quarkus.20find.20the.20deployment.20JAR.20of.20an.20extension.3F/near/416737768

@maxandersen

PR #1984, opened 2024-05-18
Add edit-this-page for SNAPSHOT

marcelstoer:
Rather than only displaying the edit-this-page link for the latest release, we also display it for the "main" version i.e. the latest SNAPSHOT.

Fixes #1878


gsmet (approved):
That makes perfect sense, thanks a lot!

Issue #1990, opened 2024-05-24
Extract texts from jekyll html templates


ynojima:
To achieve localization of the quarkus.io website, we extract texts from .adoc and .md files, save them in .po files, translate the .po files, and then write it back during jekyll site build.
However, some of the Jekyll HTML template files contain Jekyll directives, making it impossible for the text extraction parser to correctly process them. 
Currently, we achieve localization by manually replacing the entire template file with translated versions, but this method is cumbersome to follow for template file updates, resulting in poor maintainability. 
To address this HTML template file issue, we should rewrite these HTML template files loading texts from YAML files.

As of today, we need to rewrite these files:

- _includes/index-docs.html
- _includes/publication-entries.html
- _includes/recent-posts-band.html
- _layouts/guides.html

pull-request: https://github.com/quarkusio/quarkusio.github.io/pull/1991

PR #1991, opened 2024-05-24
Extract texts from templates

ynojima:
Issue: https://github.com/quarkusio/quarkusio.github.io/issues/1990

While localization repositories extracts texts from .adoc and .md files with po4a utility, po4a cannot extract texts from html template files as they contains template directives, which is not valid html tag. To localize texts in html template files, localization repositories have whole html template file instead of just extracted texts. See: https://github.com/quarkusio/ja.quarkus.io/tree/main/l10n/override/ja_JP

This impacts maintenability of lozalization repositories because once original html template files are updated, localized html template files need to be updated even when it doesn't contain texts change.

This commit extracts texts from html template files to yaml files. Once this change is introduced, localization repositories can process yaml files with po4a.


gsmet:
@ynojima sorry for the delay, I will have a look soon. I totally agree we should do this.

ynojima:
Thank you! I merged `origin/main` in 8c14ab5 to resolve conflict.

PR #1992, opened 2024-05-24
Rewrite allowfullscreen attribute xhtml valid

ynojima:
Since text extraction parser used for lozalization can parse XHTML valid mark up only, rewrite allowfullscreen attribute


gsmet (approved):

gsmet:
Thanks!

PR #1995, opened 2024-06-04
Switch from Twitter to X

gsmet:
And fix a few unrelated things.

### DIFF
--- a/_includes/socials.html
+++ b/_includes/socials.html
@@ -17,7 +17,7 @@
     <div class="width-3-12 width-0-12-m"></div>
     <div class="width-3-12 width-0-12-m"></div>
     <div class="width-6-12 width-12-12-m">
-      <a href="http://www.twitter.com/@quarkusio" class="button-cta"><i class="fab fa-twitter"></i>Follow us on Twitter</a>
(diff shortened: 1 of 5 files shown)


_includes/socials.html
```diff
@@ -17,7 +17,7 @@
     <div class="width-3-12 width-0-12-m"></div>
     <div class="width-3-12 width-0-12-m"></div>
     <div class="width-6-12 width-12-12-m">
-      <a href="http://www.twitter.com/@quarkusio" class="button-cta"><i class="fab fa-twitter"></i>Follow us on Twitter</a>
+      <a href="https://x.com/@quarkusio" class="button-cta"><i class="fab fa-twitter"></i>Follow us on X</a>
```
gsmet, line 20: Unfortunately, FA 5 doesn't support X so we would have to move to FA 6.
Not sure the update will be straightforward and we need to make sure we can still use it for free.

gsmet (commented):
@insectengine could you have a look at the comments?

gsmet:
I merged this first step, I will let @insectengine address the comments above when he has some time.

insectengine:
Thanks... looking into the switch to FA 6 to remove the need of images and pruning unused includes.

PR #2025, opened 2024-06-18
Advertise OTel support over OpenTracing in K8 native page

michalvavrik:
Kubernetes native page https://quarkus.io/kubernetes-native/ says we utilize OpenTracing but right now the priority is on OpenTelemetry.


michalvavrik:
\cc @gsmet

gsmet (approved):
Good catch, thanks!

PR #2037, opened 2024-07-02
fix: set homepage background size to 100%

ricciardelli:
The background image breaks for resolutions higher than 2352px, which is the width of the image. With this change, the background should not look broken.

Before:
Screenshot 2024-07-02 at 17 49 56

After:
Screenshot 2024-07-02 at 17 50 32

### DIFF
--- a/_sass/layouts/home.scss
+++ b/_sass/layouts/home.scss
@@ -6,6 +6,7 @@
 
   background-image: url($baseurl + '/assets/images/bg-home-primary.png');
   background-repeat: no-repeat;
+  background-size: 100%;
   p {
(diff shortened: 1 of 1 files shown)


insectengine:
This fix causes issues in mobile (this is why we changed it). Wasn't able to test at that size noted since I don't have a monitor that large. I've submitted a PR where the image was adjusted to be repeatable in the x-axis and adjusted the CSS.

https://github.com/quarkusio/quarkusio.github.io/pull/2040

maxandersen:
#2040 merged so I think this is out-of-date/duplicate now. Thanks for the pr @ricciardelli !

PR #2040, opened 2024-07-08
homepage image issue and footer issue

insectengine:
1) Make homepage image repeatable horizontally and changed css to address large screen issues ( replaces https://github.com/quarkusio/quarkusio.github.io/pull/2037)

2) fix coloring issue of icons in Red Hat footer (no visible in light mode).


_sass/layouts/home.scss
```diff
@@ -5,7 +5,8 @@
 .homepage {
 
   background-image: url($baseurl + '/assets/images/bg-home-primary.png');
-  background-repeat: no-repeat;
+  background-repeat: repeat-x;
```
cescoffier, line 8: Wouldn't `round` be better? Do you fear distortion?

insectengine: Not sure how Round would be a better solution. We have no need to repeat the image vertically and I've made the image seamlessly repeat on the X axis (while also making is smaller). This solves the issue for large screens and doesn't negatively impact mobile screens.

PR #2052, opened 2024-07-20
Fix all config page in dark mode

gsmet:
Currently, the page looks like this cause it has a mix of dark and light styles:

Screenshot from 2024-07-20 16-40-08

Apparently, this page was forced light on purpose but I'm not sure it's a good idea and I think we should revisit this decision as long pages are typically pages for which you will want to use your preferred background.

So I fixed it to use a dark background.

Ideally, it would be nice to have odd/even backgrounds even for dark mode: atm they are of the same color.

/cc @insectengine @maxandersen

### DIFF
--- a/_sass/layouts/guides.scss
+++ b/_sass/layouts/guides.scss
@@ -1,8 +1,3 @@
-
-.guides-configuration-reference {
-  background: $white;
-}
-
 div.guides,
 div.guides-configuration-reference {
   padding-bottom: 6rem;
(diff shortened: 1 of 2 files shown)


gsmet:
I merged as the current state is really bad.

insectengine:
@gsmet  - put in a PR to fix striping. 

https://github.com/quarkusio/quarkusio.github.io/pull/2055

PR #2064, opened 2024-08-01
Fix darkmode input text problem on config widget

insectengine:
Fixes a darkmode input text problem on docs "config" widgets


gsmet (approved):
Awesome, thanks!

PR #2067, opened 2024-08-02
Allow overriding toc levels at the guide level

gsmet:
For reference guides, having a deeper TOC might make sense, typically for the Qute reference guide.

### DIFF
--- a/_layouts/guides.html
+++ b/_layouts/guides.html
@@ -59,7 +59,7 @@ layout: base
       {{ content }}
     </div>
     <div class="grid__item width-4-12 width-12-12-m tocwrapper">
-      <div class="hide-mobile toc">{{ page.document | tocify_asciidoc: 2 }}</div>
+      <div class="hide-mobile toc">{{ page.document | tocify_asciidoc }}</div>
     </div>
   </div>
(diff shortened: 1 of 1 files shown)


gsmet:
Defaut is 2 so this change doesn't affect the default behavior.

_layouts/guides.html
```diff
@@ -59,7 +59,7 @@ <h1 class="text-caps">{{page.title}} {{page.docversion}}</h1>
       {{ content }}
     </div>
     <div class="grid__item width-4-12 width-12-12-m tocwrapper">
-      <div class="hide-mobile toc">{{ page.document | tocify_asciidoc: 2 }}</div>
```
mkouba, line 62: So the `: 2` is used to specify the default value, or? I mean if we remove it what's the default?

gsmet: Having a value here enforces the value and completely ignore the `:toclevels:` at the guide level.

The default for `:toclevels:` is 2 (https://docs.asciidoctor.org/asciidoc/latest/toc/levels/#set-toclevels) so by removing this, we don't change the default behavior and we allow to override it at the guide level (as I did for Qute).

PR #2079, opened 2024-08-13
Direct podman blog readers to the podman guide, which is more current

holly-cummins:
When I searched for 'podman quarkus' last week, the https://quarkus.io/blog/quarkus-devservices-testcontainers-podman/ blog was the top hit. Our official guide was number six in the latest. The blog is rather badly obseleted by now, so we don't want it to be the top hit. 

When I tried again yesterday, the guide was the top hit, which I *think* is because we pushed an update to it, and that affects the search ranking. Linking to it from the blog should also help, by 
- Improving the ranking of the guide (if it has more inbound links, it ranks higher)
- If the ranking ends up wrong, directing blog readers to the right place

### DIFF
--- a/_posts/2021-12-03-quarkus-devservices-testcontainers-podman.adoc
+++ b/_posts/2021-12-03-quarkus-devservices-testcontainers-podman.adoc
@@ -7,6 +7,8 @@ synopsis: 'Expose a Podman service for usage with Quarkus Dev Services and Testc
 author: mjurc
 ---
 
+_Update: For the latest, simpler, instructions on running Podman with Quarkus, see https://quarkus.io/guides/podman[the Quarkus Podman guide]._
+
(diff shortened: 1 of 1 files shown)


mjurc (approved):
Good thinking, lgtm

gsmet (approved):

_posts/2021-12-03-quarkus-devservices-testcontainers-podman.adoc
```diff
@@ -7,6 +7,8 @@ synopsis: 'Expose a Podman service for usage with Quarkus Dev Services and Testc
 author: mjurc
 ---
 
+_Update: For the latest, simpler, instructions on running Podman with Quarkus, see https://quarkus.io/guides/podman[the Quarkus Podman guide]._
```
gsmet, line 10: Maybe we could even go with a:

```
[TIP]
====
====
```

PR #2046, opened 2024-07-15
Add working groups page and layout

cescoffier:
- Provide the script that generate the wg.yaml
- Create a workflow that update the working group yaml periodically
- Add the wg page and layout
- A bit of CSS to make the page looks ok

PR #2080, opened 2024-08-14
Working Groups style updates

insectengine:
Fixes #2046

### DIFF
--- a/.github/workflows/sync-working-groups.yml
+++ b/.github/workflows/sync-working-groups.yml
@@ -1,0 +1,46 @@
+name: Sync working groups
+
+env:
+  JVM_VERSION: '21'
+
+on:
+  schedule:
+    - cron: '0 * * * *'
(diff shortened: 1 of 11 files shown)


gsmet:
@cescoffier let's iterate it (and don't overwrite @insectengine 's changes!)

Summary of my comments in the other PR:

- ~~we need a dark theme for the cards (and probably more love from @insectengine)~~
- you're talking about contacting a point of contact but there's none
- not completely sure about having `wg` as the url. I would have gone with `working-groups` for better SEO. You will have a link or a bookmark.
- ~~we need to include it in the nav somewhere (probably something to discuss with @insectengine )~~

In any case, nothing blocking, I let you address what you think makes sense and merge.

insectengine:
I also took the liberty to change the URL to "/working-groups" as this is consistent with our other URLs

PR #2095, opened 2024-08-23
Fix the definition of the page filename

marcelstoer:
The original implementation did not consider that `page.path` has a different pattern for the "Main - SNAPSHOT" version. This currently leads to HTTP 404 links in such cases.


marcelstoer:
@gastaldi FYI

gsmet (approved):
Awesome, thanks for taking the time to create a fix!

PR #2104, opened 2024-08-29
Pre-connect to the search service on guide search pages

yrodiere:
The first connection to the remote search service can be slow, and on these pages we're pretty sure to use the remote search service. Pre-connecting helps speed up the first call to the service.

### DIFF
--- a/_layouts/base.html
+++ b/_layouts/base.html
@@ -70,6 +70,13 @@
   <link rel="alternate" hreflang="x-default" href="https://quarkus.io/" />  
   <script src="{{ site.baseurl }}/assets/javascript/tracking.js"></script>
   {% if search_wc %}
+    {% comment %}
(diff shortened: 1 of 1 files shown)


gsmet:
Maybe this explains why not working in Firefox: https://www.reddit.com/r/firefox/comments/17c41ii/early_hints_preconnect_is_now_supported_in/

Not sure if they enabled it by default?

gsmet (approved):

yrodiere:
> Maybe this explains why not working in Firefox: https://www.reddit.com/r/firefox/comments/17c41ii/early_hints_preconnect_is_now_supported_in/
> 
> Not sure if they enabled it by default?

Yeah I saw that page and checked my about:config: it is enabled by default, at least in my case.

Really no idea what's going on.

PR #2114, opened 2024-09-12
Change reactive feature on the landing page

cescoffier:
Update the "reactive" core feature on the landing page to be more in phase with the current (and next to be) continuum page content.

### DIFF
--- a/_includes/header-navigation.html
+++ b/_includes/header-navigation.html
@@ -15,7 +15,7 @@
         <ul class="submenu">
           <li><a href="{{site.baseurl}}/about" class="{% if page.url contains '/about/' %}active{% endif %}">WHAT IS QUARKUS?</a></li>
           <li><a href="{{site.baseurl}}/container-first" class="{% if page.url contains '/container-first/' %}active{% endif %}">CONTAINER FIRST</a></li>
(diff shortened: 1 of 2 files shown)


holly-cummins:
I'm not sure "Continuum" on its own tells a reader very much about what they might find if they click on it. People wouldn't tell their boss "Yeah, we have to switch to Quarkus because it has *continuum*."

I do like the [homepage-features-icon-band.html](https://github.com/quarkusio/quarkusio.github.io/pull/2114/files#diff-4da0a5de4f1dde18369f8d653effba4ea8e1389ad9e7b0bd9b1ce1ecf9f74b2d) changes, though.

holly-cummins:
> I agree "continuum" is not great. But it's way better than "reactive" which point to a page barely mentioning reactive.

You're right, this is hard. I never thought to look at the link, and now that I've done it, I'm not entirely sure what the page is about. So I agree, "Reactive" is not right and should be changed. 

> What about versatility?

+1. I think that does a better job of capturing the user-facing value. (And also of summarising what that page is trying to say.)

cescoffier:
@holly-cummins - wording changed in the menu.

holly-cummins (approved):

holly-cummins:
Will mirror to http://quarkus.io/extensions once this merges.

PR #2117, opened 2024-09-13
Update link which has been renamed on the wiki

holly-cummins:
It is now https://github.com/quarkusio/quarkus/wiki/Migration-Guide-3.0-Hibernate-ORM-5-to-6-migration


holly-cummins:
I realised that the file is references in the guides for several LTS versions (https://github.com/holly-cummins/quarkus/tree/update-renamed-link), so the rename change will be pretty invasive in terms of backports. I will try to work around the problem name and undo the rename instead.

PR #2135, opened 2024-10-15
Csp adobe analytics adjustment to fix data collection issue

insectengine:


cescoffier (approved):
A bit worried that we do not have SHA anymore which lower the security, but it should be ok.

_includes/head-csp.html
```diff
@@ -0,0 +1,24 @@
+<meta http-equiv="Content-Security-Policy" content="
+  default-src https://dpm.demdex.net https://route-default-test-mscherer-matamo.apps.ospo-osci.z3b1.p1.openshiftapps.com/ {{ site.search.host }}; 
+  script-src 'self' 'unsafe-inline' 'unsafe-eval'
+        {{ search_script }} 
+        https://assets.adobedtm.com
```
maxandersen, line 5: do we really need to remove shas for *everything* and not just the adobe assets?

insectengine: Unfortunately, I've tested it by adding the sha I knew was associated with the light/dark mode and it broke the adobe collection.

maxandersen:
not great, but at least its over https. we might get flagged with "lower" scoring on some security site reviews but in practice this seems not a big thing to worry about. merging.

PR #2153, opened 2024-11-04
Remove the 2.7 guides data yml

marko-bekhta:
A follow-up on https://github.com/quarkusio/quarkusio.github.io/pull/2152

since the 2.7 guides were removed the data file should also go away (I noticed it since the search service "discovered" the yml but couldn't find any HTML guides 🙈 😃 )


yrodiere:
LGTM, also opened this: https://github.com/quarkusio/search.quarkus.io/issues/351

PR #2157, opened 2024-11-09
Add inline macro for copying of configuration properties in dotted format

michalvavrik:
This new inline macro is required for https://github.com/quarkusio/quarkus/pull/44403.


gastaldi (approved):
Nice!

PR #2168, opened 2024-11-18
Update FAQ with links to extensions.quarkus.io and syntax updates

holly-cummins:
I was trying to google for the maturity matrix, but it's not indexed yet. What google did send me to was the general FAQ, which has a line about how we have a maturity model, and no link. There couldn't be a link because the reference is several years old. I can't add a link until the matrix doc is in a 'latest' release, but while I was looking at the doc, I noticed a few other areas that could use some love. 

There's a suggestion to use code.quarkus.io to check extension status, but extensions.quarkus.io is a more natural place to look up that kind of metadata. It also assumes no extensions live in the quarkiverse org, presumably because it was written before there was one. 

The wording about stability is a bit weak, so I strengthened with a link to our user stories. 

I also noticed the headings were markdown ones, so I swapped them to adoc style.

Finally, there was a link to the 'orginal' /extensions page, which now redirects to code.quarkus.io. For that link, code.quarkus.io is a much better destination, so I've replaced /extensions with code.quarkus.io.

### DIFF
--- a/faq.adoc
+++ b/faq.adoc
@@ -11,13 +11,15 @@ Quarkus is an Open Source project licensed under the https://www.apache.org/lice
 
 == Where can I get it?
 
-Quarkus is published in Maven Central, check out link:/extensions[which extensions] you need and just import them in your `pom.xml` to get Quarkus. We recommend you start your Quarkus experience via our link:/get-started[Getting Started guides].
+Quarkus is published in Maven Central. We recommend you start your Quarkus experience with our link:/get-started[Getting Started guides] or by downloading a scaffolded application from https://code.quarkus.io[code.quarkus.io].
 
 == Quarkus is stable?
 
-Yes, we consider Quarkus stable. Remember 95% of the features Quarkus apps use are provided by the ecosystem like Hibernate ORM, Eclipse Vert.x, Netty, RESTEasy, etc. These libraries are rock solid :)
+Yes, we consider Quarkus stable.
+Quarkus is used in production by a wide range of organizations (see some link:blog/tag/user-story/[user stories]).
+Remember 95% of the features Quarkus apps use are provided by the ecosystem like Hibernate ORM, Eclipse Vert.x, Netty, RESTEasy, etc. These libraries are rock solid :)
 
-You can go to https://code.quarkus.io/[code.quarkus.io] to see if a particular extension is _stable_ (default) or _preview_.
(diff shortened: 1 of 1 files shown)


faq.adoc
```diff
@@ -11,13 +11,15 @@ Quarkus is an Open Source project licensed under the https://www.apache.org/lice
...
 
-Quarkus is published in Maven Central, check out link:/extensions[which extensions] you need and just import them in your `pom.xml` to get Quarkus. We recommend you start your Quarkus experience via our link:/get-started[Getting Started guides].
+Quarkus is published in Maven Central. We recommend you start your Quarkus experience via our link:/get-started[Getting Started guides] or by downloading a scaffolded application from http://code.quarkus.io[code.quarkus.io].
 
 == Quarkus is stable?
 
-Yes, we consider Quarkus stable. Remember 95% of the features Quarkus apps use are provided by the ecosystem like Hibernate ORM, Eclipse Vert.x, Netty, RESTEasy, etc. These libraries are rock solid :)
+Yes, we consider Quarkus stable.
+Quarkus is used in production by a https://quarkus.io/blog/tag/user-story/[range of organisations].
+Remember 95% of the features Quarkus apps use are provided by the ecosystem like Hibernate ORM, Eclipse Vert.x, Netty, RESTEasy, etc. These libraries are rock solid :)
```
gsmet, line 20: I'm not sure about this content now tbh but probably not something to discuss in this PR.

I wonder if I would have `wide range of organizations (see some link:user stories)` because otherwise it's a bit limiting. But not really sure and also don't want to start a long discussion here.

gsmet (commented):
I added some comments, the only one I really want addressed is to avoid the http link. For the rest, I  let you judge of what is best.

faq.adoc
holly-cummins: Yeah, I decided to go for "make it better than it currently is" rather than "make it optimal". I agree a bigger rewrite is needed.

holly-cummins: I don't have strong feelings about either wording, so I've gone with your suggestion and squashed.

holly-cummins:
Could I nudge on this? I've got another change backed up behind it ...

gsmet:
Sorry, missed your ping, I fixed the conflict for you because the issue came from me lagging.

I'll merge once CI is green.

holly-cummins:
Thanks!

PR #2175, opened 2024-11-26
Add link for maturity model to faq paragraph which references it

holly-cummins:
Our FAQ document references a maturity model, but there's no backing link. This because the reference is many years old, and we only wrote the maturity model down this year. The FAQ is one of the top hits for 'quarkus maturity model', so it really should have the right link on it.


holly-cummins:
Preview link: https://quarkus-site-pr-2175-preview.surge.sh/faq/

gsmet (approved):

PR #2178, opened 2024-11-27
Dev UI section and page

josephaw1022:
Adding a Dev UI Page and a section to the home page

PR #2179, opened 2024-11-28
Reorder 'features' on front page to emphasise performance

holly-cummins:
The current order has developer joy and performance as the least prominent benefits, which isn't ideal:

image

I've also updated the order of the navbar menus to match the new order. (It didn't quite match the old order, either. :) )

### DIFF
--- a/_includes/homepage-features-icon-band.html
+++ b/_includes/homepage-features-icon-band.html
@@ -6,28 +6,10 @@
   </div>
   <div class="grid-container">
     <div class="width-4-12 width-12-12-m contrib-block">
-      <img class="light-only" src="{{site.baseurl}}/assets/images/about/icon-containerfirst.svg">
-      <img class="dark-only" src="{{site.baseurl}}/assets/images/about/icon-containerfirst-dark.svg">
-      <h3><a href="/container-first">Container First</a></h3>
-      <p>Quarkus tailors your application for GraalVM and HotSpot. Amazingly fast boot time, incredibly low RSS memory (not just heap size!) offering near instant scale up and high density memory utilization in container orchestration platforms like Kubernetes. We use a technique we call compile time boot.</p>
-    </div>
-    <div class="width-4-12 width-12-12-m contrib-block">
-      <img class="light-only" src="{{site.baseurl}}/assets/images/about/icon-reactive.svg">
-      <img class="dark-only" src="{{site.baseurl}}/assets/images/about/icon-reactive-dark.svg">
-      <h3><a href="/continuum">Reactive Core</a></h3>
-      <p>Built on a robust reactive core, Quarkus ensures fast and efficient performance, supporting the development of a wide variety of modern applications.</p>
-    </div>
-    <div class="width-4-12 width-12-12-m contrib-block">
-      <img class="light-only" src="{{site.baseurl}}/assets/images/about/icon-standards.svg">
-      <img class="dark-only" src="{{site.baseurl}}/assets/images/about/icon-standards-dark.svg">
-      <h3><a href="/standards">Community and Standards</a></h3>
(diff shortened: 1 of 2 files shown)


holly-cummins:
New order: 
image

I tried to line things up vertically, too, so 'reactive core' supports 'performance'. I perhaps should have carried that on and had 'standards' supporting 'developer joy' and 'container first' supporting 'kube native'.

holly-cummins:
Revision. Now there's (very very loose) 'supports/enables' semantics on each column: 
image

_includes/homepage-features-icon-band.html
```diff
@@ -6,10 +6,22 @@ <h2>Quarkus Benefits</h2>
...
     <div class="width-4-12 width-12-12-m contrib-block">
-      <img class="light-only" src="{{site.baseurl}}/assets/images/about/icon-containerfirst.svg">
-      <img class="dark-only" src="{{site.baseurl}}/assets/images/about/icon-containerfirst-dark.svg">
-      <h3><a href="/container-first">Container First</a></h3>
-      <p>Quarkus tailors your application for GraalVM and HotSpot. Amazingly fast boot time, incredibly low RSS memory (not just heap size!) offering near instant scale up and high density memory utilization in container orchestration platforms like Kubernetes. We use a technique we call compile time boot.</p>
+      <img class="light-only" src="{{site.baseurl}}/assets/images/performance/icon-performance.svg">
+      <img class="dark-only" src="{{site.baseurl}}/assets/images/performance/icon-performance-dark.svg">
+      <h3><a href="/performance">Performance</a></h3>
+      <p>Quarkus streamlines framework optimizations in the build phase to reduce runtime dependencies and improve efficiency. By precomputing metadata and optimizing class loading, it ensures fast startup times for JVM and native binary deployments, cutting down on memory usage.</p>
```
cescoffier, line 12: What's the reason to put performance before developer joy?

cescoffier:
More I read kubernetes and Container more I think they should be somewhat merged.

_includes/homepage-features-icon-band.html
holly-cummins: No good reason. I agree the other way round is probably better, will swap.

insectengine:
Sorry, have been out on holiday last week.

Keep in mind that the page is responsive and there isn't an intent to have columns of content as in mobile, it all turns to a single column. We also have the DevUI content coming up for consideration. I mentioned there that that content feels like it should be added to this band (and not have it's own band). This would mean we'll have 7 bits of content to order, unless we combine the Container/Kube page and add a redirect for the abandoned page. I agree that the Container/Kube pages can be combined.

holly-cummins:
> Sorry, have been out on holiday last week.
> 
> Keep in mind that the page is responsive and there isn't an intent to have columns of content as in mobile, it all turns to a single column. We also have the DevUI content coming up for consideration. I mentioned there that that content feels like it should be added to this band (and not have it's own band). This would mean we'll have 7 bits of content to order, unless we combine the Container/Kube page and add a redirect for the abandoned page. I agree that the Container/Kube pages can be combined.

Let's separate out the combining of container and kube from the problem this PR is solving, which is "do not have performance and developer joy be the least important things in the list." I agree that if we add dev ui, that would be a good opportunity to combine them, to avoid going to seven benefits (which is too many to quickly take in).

maxandersen:
+1 on just having this reoroderd to have devjoy higher.

devui I'm not sure make sense to have in this but lets have that in separate pr/issue if we got some reasons for it.

holly-cummins:
> devui I'm not sure make sense to have in this but lets have that in separate pr/issue if we got some reasons for it.

The Dev UI work is heppening in this PR: https://github.com/quarkusio/quarkusio.github.io/pull/2178

PR #2193, opened 2024-12-12
fix: workaround jekyll tags bug

maxandersen:
a bug from 2014 (https://github.com/jekyll/jekyll-archives/pull/24) result in when tags has mixed casing the archived list varies dependent on the order of tags.

so basically jekyll messes up tag listing if tags names are mixed casesr
for ai we have "ai", "ai", "AI", "ai" in sequence. the AI overrides the two first, the last ai overrides "AI"
so its some funky hashcollision thing

I wrote a script to spot which posts had mixed cases - seems we are pretty good at it already :)

also cleaned up to have consistent format of tags and frontmatters.

all just so https://quarkus.io/blog/tag/ai and https://quarkus.io/blog/tag/openshift will return consistent result.

### DIFF
--- a/_posts/2019-09-23-quarkus-developer-joy-for-vs-code.adoc
+++ b/_posts/2019-09-23-quarkus-developer-joy-for-vs-code.adoc
@@ -1,7 +1,7 @@
 ---
 layout: post
 title: 'Quarkus developer joy for VS Code'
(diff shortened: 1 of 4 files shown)


maxandersen:
pushed wrong branch - replaced with #2194

PR #2198, opened 2024-12-17
Quarkus and Raspberry Pi - bare metal - blog fix

mkouba:
- in the "Run in production" a slow startup is mentioned, however it should be only related to the dev mode


franz1981 (approved):
Well done!

PR #2201, opened 2024-12-18
Use the new tracker url

mscherer:


insectengine:
combined with other change in this PR: https://github.com/quarkusio/quarkusio.github.io/pull/2202

Issue #2211, opened 2025-01-20
Build user story landing page, with instructions for submitting, and link it more prominently on the web page


holly-cummins:
I was trying to give someone instructions on how to submit a user story blog, and realised I couldn't link to anything. 

It would be nice if https://quarkus.io/blog/tag/user-story/ had a section at the top that said "Do you have a user story? We'd love to publish it. Here's how to share your story." or something like that.

This might be easiest with a dedicated landing page rather than a blog tag. That page could also use some other terms people might search for, like "references" and "success stories". 

It would also be good to link to the user story page more prominently on the front page, perhaps in the footer.

insectengine:
> It would be nice if https://quarkus.io/blog/tag/user-story/ had a section at the top that said "Do you have a user story? We'd love to publish it. Here's how to share your story." or something like that.

Can you flesh out the instructions for submitting a use case?

insectengine:
https://github.com/quarkusio/quarkusio.github.io/pull/2227#issuecomment-2635369818

PR #2227, opened 2025-02-04
Add Basic User stories page 

insectengine:
This adds the basic user stories page with link in primary nav flyout and footer. Once we have a landing page/post for how to contribute a user story, we'll add an additional band with a CTA to direct to that new content.


insectengine:
Partially solves #2211

PR #2258, opened 2025-03-14
Fixed 'runtime' typo to 'build-time' on Performance page

paulrobinson:
I assume 'build-time' is meant in this context...


gsmet (approved):
Ah ah, yes, this is a bad one. Good catch!

PR #2271, opened 2025-03-30
Fix template generating invalid wg.yaml

ynojima:
With this fix, YAML remains valid even when commas are included in the short-description like this:

https://github.com/quarkusio/quarkusio.github.io/blob/fb1bdb9a0d89afec8cf3d1d77c6f509bec79f6f0/_data/wg.yaml#L24


jmartisk:
Thanks!

PR #2275, opened 2025-04-01
Add 3.20 docs, prune old versions

jmartisk:


_data/versions.yaml
```diff
@@ -7,8 +7,7 @@ maven: 3.9.6+
 documentation:
   - 'main'
   - 'latest'
+  - '3.20'
   - '3.15'
   - '3.8'
   - '3.2'
-  - '2.16'
```
maxandersen, line 13: we shuoldn't do this if it means the guides urls gets removed....

maxandersen: I suggest to split this in two - one to get 3.20 added ASAP
and if need to separate one figuring out how we handle archiving so we don't break doc links. we can there look at access data/stats to see how often they are accesses.

jmartisk: Ok, I've put back the old versions..

gsmet (approved):

PR #2292, opened 2025-04-30
Fixed typos in Kotlin Live Reload

skyrylyuk:
Kotlin compiler has name 'kotlinc' instead of 'kotlin'
Method 'compiler().args()' has parameters type 'List' instead of 'Array'


gsmet:
Thanks @skyrylyuk . Any chance you could create a similar PR for our main repository? The doc is copied from there.

See https://github.com/quarkusio/quarkus/blob/main/docs/src/main/asciidoc/kotlin.adoc .

Thanks!

PR #2294, opened 2025-05-02
Add cautions about newly reported @Nested regressions to classloading blog

holly-cummins:
A few reports have come in that might affect Gradle users, so we should list them. 
Direct preview link: https://quarkus-site-pr-2294-preview.surge.sh/blog/test-classloading-rewrite/


holly-cummins:
Ah, pause, one of these issues isn't a gradle thing, it's an accessing config without using `@QuarkusTest` thing.

gsmet (approved):

PR #2356, opened 2025-07-16
add CF band to homepage and CF footer

insectengine:
Contained in this commit:
1) Added home page Commonhaus Foundation band (with dark and light mode support).
2) Modified the footer band to accommodate CF requirements (logo/disclaimer text).
3) Moved creative commons icons with links and the privacy policy links from the black bar to the subproject band.
4) Refactored the media breaks and styling for the subproject band to make it a cleaner/visually compelling responsive design.


_data/projectfooter.yaml
```diff
@@ -28,6 +28,8 @@ links:
         url: /brand
       - page: Wallpapers
         url: /desktopwallpapers
+      - page: Privacy Policy
+        url: https://www.redhat.com/en/about/privacy-policy
```
ebullient, line 32: Does the Red Hat privacy policy still apply? Commonhaus has one for itself (but not necessarily for its projects): https://www.commonhaus.org/policies/privacy-policy/

insectengine: The privacy policy is required since we're still using Adobe Analytics. This covers us for GDPR compliance.

holly-cummins:
On my screen (which is large, but not redonculously large), the sizing seems a bit off, relative to the text in the other bands:

image

Also, I wonder if the text next to the logo would look better if it was vertically centred, so it was less out of alignment with the text in the logo?

insectengine:
@holly-cummins - We use a 12 column grid and that version it was a 7/5 split of logo to content. I've reversed it to 5/7 with my last commit.

Screenshot 2025-07-17 at 2 20 11 PM
Screenshot 2025-07-17 at 2 22 36 PM

Sanne:
It looks great and I'm eager to merge it - would it be ok if I pressed the big green button?

I do have some minor nitpicks but I think I'd prefer to let @insectengine to choose if he'd want to handle them as a follow up or not at all :)

Here is my screenshot:

image

Two things stand out:
 - the commonhaus logo in the footer seems rather big - perhaps it should be of similar height of the Quarkus logo which is in the center/top of the picture?
 - the "Sponsored by" lettering seems to drift a bit far from the Red Hat logo on such a display

Both of these look totally fine on mobile though. I appreciate my display is large `3360x1440` - on the other hand I expect most developers to have large screens, often even larger.

Neither seem blocking so I'd rather merge and polish later though.  It's also still undecided if we should keep the "Sponsored by Red Hat" altogether so wouldn't want you to spend too many cycles on it, if it's not a trivial change.

Sanne (approved):

jponge:
Thanks I'll transition that to the Mutiny projects, and future releases will reflect that change.

Sanne:
Good point, but bear in mind that there's the possibility that @insectengine will want to apply some more tweaks - perhaps give it some days for things to settle, so to avoid a multitude of patches.

insectengine:
@Sanne - Think this issue will require a change of how this is built as columns based layout will force this space. We can limit the size of the CF logo, but that will just emphasize the spacing problem. I'll have to refactor this a bit #bigscreenproblems

@jponge @radcortez  @dmlloyd  - hold off on this until I get this sorted.

PR #2358, opened 2025-07-18
refactored CF bar to use flexbox instead of columns. This should clea…

insectengine:
This should clear up issues on ultra large screens but I have no way to test this locally.

### DIFF
--- a/_includes/CF-footerband.html
+++ b/_includes/CF-footerband.html
@@ -1,16 +1,18 @@
 <div class="content cf-footer">
-  <div class="grid-wrapper">
+  <div class="flexcontainer">
     <span class="cf-logo">
       <img src="https://raw.githubusercontent.com/commonhaus/artwork/main/foundation/brand/svg/CF_logo_horizontal_single_reverse.svg"/>
     </span>
     <span class="license">
       Copyright © Quarkus. All rights reserved. For details on our trademarks, please visit our <a href="https://www.commonhaus.org/policies/trademark-policy/">Trademark Policy</a> and <a href="https://www.commonhaus.org/trademarks/">Trademark List</a>. Trademarks of third parties are owned by their respective holders and their mention here does not suggest any endorsement or association.
     </span>
-    <span class="sponsor">
-      Sponsored by
-    </span>
-    <span class="sponsor-logo">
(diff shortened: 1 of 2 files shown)


insectengine:
Some screenshots of the footer at various points of repsonsiveness.

Screenshot 2025-07-18 at 2 40 24 PM
Screenshot 2025-07-18 at 2 40 52 PM
Screenshot 2025-07-18 at 2 40 33 PM

On the Project sub-footer (the part with desc, links and such). The quarkus horizontal logo is intended to be centered.
Screenshot 2025-07-18 at 3 44 10 PM


In general, for desktop is designed around looking the best at 1920px wide, the site is designed to have 13rem of padding on the left and right (it's considerably smaller on mobile and tablet). This means, if you're in mega theatrical wide wide mode of 3500px, it's going to look a bit... stretched out.

One option is to look to fix the width of the content at 1920. So at larger sizes you'll get a centered site with extra padding on the sides. We used to have the site fixed but I think folks wanted it to fill. Not sure what the answer is here.

Sanne:
@insectengine the screenshots you shared look fine - I would agree that even while many might have super-wide screens, we don't usually show the browser on full-screen so don't spend too much time on this.

>One option is to look to fix the width of the content at 1920. So at larger sizes you'll get a centered site with extra padding on the sides. We used to have the site fixed but I think folks wanted it to fill.

Personally I think it would be fine to cap the width in case of it getting just too much, but I wonder what happens on high dpi screens? I don't have one of those - is that possibly why some folks wanted it to fill?
If that's not the case I'd suggest don't try to make everyone happy, you're the expert and can decide to make some tradeoffs; personally I think what you have above is beatiful in all setups we care about.

holly-cummins:
I haven't looked at the changes yet, but last week I was thinking "should I suggest to @insectengine that flexbox makes everything better? No, Holly, don't do that, James has lots of things to be doing without refactoring column layouts away just because they're not the current fashion ... "  

So I'm happy to see this change arrived without me even having to irritate anyone  :)

Sanne:
Should we merge this?

holly-cummins:
> Should we merge this?

Just looking now!

holly-cummins (approved):

PR #2364, opened 2025-07-22
Make the commonhaus logo in the footer a hyperlink

holly-cummins:
The Red Hat logo is a hyperlink, but not the CommonHaus one. Should they both be?


insectengine:
Sorry about that, seems during the refactoring from columns to flexbox, that got dropped.

PR #2365, opened 2025-07-22
Center cf logo in footer, vertically

holly-cummins:
Before: 

image

image

After:

image

I think the main thing that made a difference was scattering `display:flex` into the anchors wrapping the images. I also switched from spans to divs, since I think that's closer to what's wanted, semantically, and also used the `gap` feature of flexbox instead of padding the individual containers asymmetrically. 

image


insectengine (approved):
Looks good!

PR #2394, opened 2025-09-09
Remove some sections from preview to fit under the surge limit

holly-cummins:
This won't have a successful preview on the PR request, since it uses `main`'s script, but hopefully after merging previews will start working again. 

Sadly, I can't test I've not broken things, for the same reason. :)


holly-cummins:
Darn. I wonder if I got the script wrong, because the next preview after this still failed with 'application too large,' and I should have shrunk things by at least 50mb. Either that or surge have reduced the limit from 400mb.

PR #2423, opened 2025-09-29
Add a GitHub workflow to update the Quarkus Insights data file

### LINKED ISSUE #2342: Ensure past-present status for Quarkus Insights episodes is always correct
The most likely time someone might try to find the link for a Quarkus Insights episode is shortly after its happened, because it's fresh in their mind and they want to share it. Unfortunately, this is also the time when the page is most likely to be incorrect, because we rely on a manual code change and rebuild to update the page. 

For example, I'm trying to share today's episode on social media, and I can't. 90m after it aired, it's still showing as 'scheduled', with no link: 

Image

This isn't a criticism of anyone for not being fast enough – doing updates instantaneously after airing isn't a reasonable SLA for people who have meetings, PTO, other tasks, and so on. However, it's something that automation could achieve quite easily, either with on-page javascript or some backend update-and-build scripting. I know we've discussed this before but I do think it's something we should do. It would avoid showing users wrong data, and reduce the maintenance effort for @insectengine.

marko-bekhta:
* Fixes https://github.com/quarkusio/quarkusio.github.io/issues/2342

Hey 👋🏻 😃 

As I could only find APIs that require a key/token to make requests to YouTube (to find a link for a streamed video), I thought that adding some JS on the page to pull the data wouldn't fly 😃. And instead, here's a workflow that will run hourly on Mondays, polling the YouTube api to see if a video with the "next #" can be found, and if so, open a pull request to update the data file. I didn't want to go with the option to push directly to the branch, as I thought it would be best to run this for some time to make sure it applies the changes correctly.

Script requires an API token. Probably best to create a dedicated YouTube user for that. Als,o a better git user/email would be needed if we go with this 😃.

And here's an example of a PR it should open: 
- https://github.com/marko-bekhta/quarkusio.github.io/pull/3

### DIFF
--- a/.github/workflows/update-insights.yml
+++ b/.github/workflows/update-insights.yml
@@ -1,0 +1,136 @@
+name: Update Insights Sessions archive/upcoming
+
+on:
+  schedule:
+    - cron: '0 * * * 1'
+  workflow_dispatch:
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  build:
+    if: github.repository == 'quarkusio/quarkusio.github.io'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+
+      - name: Archive current session
+        id: archive_session
+        env:
+          YOUTUBE_API_KEY: ${{ secrets.INSIGHTS_YOUTUBE_API_KEY }}
+          YOUTUBE_CHANNEL_ID: ${{ secrets.INSIGHTS_YOUTUBE_CHANNEL_ID }}
+        run: |
+          today=$(date +"%B %d, %Y")
+          
+          nextsessiondate=$(yq -r '.nextsessiondate' _data/insights-videos.yaml)
+          nextsessiondate_formatted=$(date --date=$nextsessiondate +"%B %d, %Y")
+          
+          if [ "$today" != "$nextsessiondate_formatted" ]; then
+            echo "Next session date ($nextsessiondate_formatted) does not match today's date ($today). Skipping workflow run."
+            exit 0
+          fi
+          
+          current_episode=$(yq -r '.nextsessiontitle | sub(".* #([0-9]+):.*"; "${1}")' _data/insights-videos.yaml)
+          echo $current_episode
+          
+          youtube_response=$(curl -s --fail \
+              -H "X-goog-api-key: ${YOUTUBE_API_KEY}" \
+              "https://www.googleapis.com/youtube/v3/search?part=snippet&channelId=${YOUTUBE_CHANNEL_ID}&q=%23${current_episode}&type=video")
+          
+          video_id=$(echo $youtube_response | jq -r '.items[0].id.videoId')
+          video_status=$(echo $youtube_response | jq -r '.items[0].snippet.liveBroadcastContent')
+          video_title=$(echo $youtube_response | jq -r '.items[0].snippet.title')
+          
+          # Check if a video ID was found
+          if [ -z "$video_id" ] || [ "$video_id" == "null" ]; then
+            echo "Could not find the YouTube video for episode #$current_episode. Assuming it's not there yet. Exiting..."
+            exit 0
+          fi
+          
+          if [[ "$video_title" == *"#$current_episode"* ]]; then
+            echo "The video title contains the required episode number: '#$current_episode'"
+          else
+            echo "The video title does NOT contain the required episode number: '#$current_episode'. Exiting..."
+            exit 1
+          fi
+          
+          if [ "$video_status" != "none" ]; then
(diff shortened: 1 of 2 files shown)


holly-cummins:
@insectengine, what do you think of this? It would make your life easier, right?

holly-cummins:
@marko-bekhta could you update this so the spacing matches the current spacing, to make the diff easier to read?

marko-bekhta:
ah it's been a while 🙂 
that yaml file was reprocessed with `yq`, I'm not sure it allows much flexibility as to how it formats the output... I could revert it, but on the first run of the workflow, it would run the `yq`, and that will change the formatting, so at the time I was thinking -- let's run the `yq` now and then subsequent PRs from the workflow would show cleaner diffs ... 
but yeah, the yaml in this PR  is waaay outdated ...

holly-cummins:
> ah it's been a while 🙂 that yaml file was reprocessed with `yq`, I'm not sure it allows much flexibility as to how it formats the output... I could revert it, but on the first run of the workflow, it would run the `yq`, and that will change the 

I had a similar thing recently and I was able to change the indentation settings in my yaml processor to match the current settings. You're right that it only matters for the first run, so not worth spending a lot of time on, but a Claude or a Bob might be able to tweak the `yq` settings easily. (Mine was snakeyaml, I think, so not exactly the same, and different knobs.)

I agree definitely don't change it for the PR but then leave the sync job to change it next time. 

The other thing you could pick up from https://github.com/quarkusio/quarkusio.github.io/pull/2818 is the idea of doing a dry run on PRs to the workflow file. That makes it easy to evaluate the effect of changes, without doing anything terrible like writing to `main` from a PR. 

We need @insectengine's agreement, but I *do* think we should do this, because having James update the file every week isn't the best use of his time. If it was me I would have done javascript so things updated instantly after airing, but since we have so many data files being updated by sync jobs, this static pattern is actually more idiomatic.

insectengine:
Correct me if I'm wrong but this looks like it only moves the session from the "upcoming podcasts" section to the "archived podcast video" section. It doesn't update the content in the "Upcoming Podcast" hero section with the embedded video and future episode details on the right. 

Am I correct in how I'm reading the change on the yaml file?

holly-cummins:
> Correct me if I'm wrong but this looks like it only moves the session from the "upcoming podcasts" section to the "archived podcast video" section. It doesn't update the content in the "Upcoming Podcast" hero section with the embedded video and future episode details on the right.
> 
> Am I correct in how I'm reading the change on the yaml file?

I've only read the script casually, but I *think* it's doing things with the next episodes too. But @marko-bekhta can confirm. This is, of course, a good argument for adding the `--dry-run` option – it would make it easier to for us to review what actually changes, without guessing. (As I implemented my dry-run, it wouldn't change the rendered site, and we'd have to read the CI logs to see the changes, so it's a bit limited, but it's something. Or at least, I found it useful, but I did paste it into the PRs on my PRs because otherwise it was buried in the logs.)

holly-cummins:
I've made a few updates to this: 

- Use a direct push rather than a PR – scarier, but it means the update doesn't get blocked, and it's less work for us
- Added a dry run mode, so we can see the behaviour in PRs
- Switched to using YouTube RSS feed instead of YouTube API, to avoid any hassle with keys 
- Used `Sync` in the workflow name for consistency with all our other data sync workflows

Here's an example of the diff. The first run removes all the blank lines, so it's a bigger diff, but after that it should calm down. 

```
DRY-RUN: Changes that would be committed:
diff --git a/_data/insights-videos.yaml b/_data/insights-videos.yaml
index 2723fd5..d7dcf73 100644
--- a/_data/insights-videos.yaml
+++ b/_data/insights-videos.yaml
@@ -3,37 +3,25 @@
 sectiontitle: "Upcoming Podcast"
 socialdescription: "View the Podcast on these platforms"
 futureheadline: "Scheduled Podcasts"
-
 # Next upcoming session (update weekly)
-nextsessiontitle: "Episode #256: What is Floci?"
-nextsessiondate: "2026-08-10T13:00Z"
-nextsessionguest: "Hector Ventura"
-
-
+nextsessiontitle: "No Session - Summer Holiday"
+nextsessiondate: "2026-08-17T13:00Z"
+nextsessionguest: "None"
 # upcoming session list (update weekly)
 # Note: First entry should match nextsessiontitle above
 futurevideos:
-
-- title: "Episode #256: What is Floci?"
-  link:
-  date: "August 10th, 2026"
-  authors: "Hector Ventura"
-
 - title: "No Session - Summer Holiday"
   link:
   date: "August 17th, 2026"
   authors: "None"
-
 - title: "Episode #257: Taming the JVM: Optimizing Java Workloads on OpenShift & Kubernetes"
   link:
   date: "August 24th, 2026"
   authors: "Rob Sedor"
```

holly-cummins (approved):
Approving this, which is mildly dodgy since I wrote some of it, but @insectengine has looked at the diff and agreed it's looking ok.

PR #2508, opened 2026-01-08
Add labels to guides with extension status (if available)

marko-bekhta:
we'll need to coordinate merging this patch with the deployment of

- https://github.com/quarkusio/search.quarkus.io/pull/535

since `main.js` changed to `app.js` with the bundler update.

PR #2519, opened 2026-01-20
Include the extension status explanation and use status derived from extensions if not defined explicitly

marko-bekhta:
- Follows up on https://github.com/quarkusio/quarkusio.github.io/pull/2508
- Requires https://github.com/quarkusio/quarkus/pull/52100

This updated asciidoc extension would inject the status label and explanation next to it, so we wouldn't need to remember to include the `extension-status.adoc` include and make things a bit more "automated"

A few previews: 
image
image

cc: @yrodiere @FroMage @insectengine  @rolfedh

### DIFF
--- a/_plugins/asciidoctor-extension.rb
+++ b/_plugins/asciidoctor-extension.rb
@@ -163,23 +163,33 @@ Extensions.register do
   tree_processor do
     process do |doc|
       status = doc.attr('extension-status')
+      
+      if status.nil? || status.empty?
+        extension_statuses = doc.attr('extensions', '').split(',')
+            .map(&:strip)
+            .reject(&:empty?)
+            .map { |s| s.gsub(/[.:]/, '-') + '-extension-status' }
+            .map { |s| doc.attr(s, '') }
+        status = ['experimental', 'preview', 'deprecated'].find { |s| extension_statuses.include?(s) }
+      end
 
       if status && !status.empty?
-        tooltip = case status
+        explanation_inner_html = case status
           when 'experimental'
-            'This extension requests early feedback to mature the idea'
+            '<p>The extension(s) discussed here request early feedback to mature the idea.</p><p>There is no guarantee of stability nor long term presence in the platform until the solution matures. Feedback is welcome on our <a href="https://groups.google.com/d/forum/quarkus-dev">mailing list</a> or as issues in our <a href="https://github.com/quarkusio/quarkus/issues">GitHub issue tracker</a></p>.'
           when 'preview'
-            'This extension\'s backward compatibility and presence in the ecosystem is not guaranteed'
+            '<p>The backward compatibility and presence in the ecosystem of extension(s) discussed here is not guaranteed.</p><p>Specific improvements might require changing configuration or APIs, and plans to become <em>stable</em> are under way. Feedback is welcome on our <a href="https://groups.google.com/d/forum/quarkus-dev">mailing list</a> or as issues in our <a href="https://github.com/quarkusio/quarkus/issues">GitHub issue tracker</a>.</p>'
           when 'stable'
-            'This extension\'s backward compatibility and presence in the ecosystem are taken very seriously'
+            '<p>The backward compatibility and presence in the ecosystem of the extension(s) discussed here are taken very seriously.</p>'
           when 'deprecated'
-            'This extension is likely to be replaced or removed in a future version'
+            '<p>The extension(s) discussed here is likely to be replaced or removed in a future version.</p>'
           else
             ""
           end
-        label_html = %(<a class="status-label status-#{status}" title="#{tooltip}" href="#extension-status-note">#{status}</a>)
-
-        label_block = create_pass_block doc, label_html, {}
+        label_html = %(<div class="extension-status-icon"><span class="status-label status-#{status}">#{status}</span></div>)
+        explanation_inner_html = %(<div class="extension-status-explanation">#{explanation_inner_html}For a full list of possible statuses, check our <a href="/faq/#what-are-the-extension-statuses">FAQ entry</a>.</div>)
+        staus_block_html = %(<div class="extension-status">#{label_html}#{explanation_inner_html}</div>)
+        label_block = create_pass_block doc, staus_block_html, {}
         doc.blocks.insert(0, label_block)
       end
       doc
--- a/_sass/asciidoc.scss
+++ b/_sass/asciidoc.scss
@@ -149,10 +149,8 @@ pre.highlight {
   line-height: 1em;
   text-transform: uppercase;
   font-weight: bold;
-  text-decoration: none;
   display: inline-block;
   padding: 4px 12px;
-  margin-bottom: 35px;
   border-radius: 50px;
 }
 
@@ -161,17 +159,9 @@ pre.highlight {
   background-color: var(--tag-stable-background-color);
 }
 
-a.status-stable:hover, a.status-stable:active, a.status-stable:focus {
-  color: var(--tag-stable-text-color);
-}
-
 .status-preview {
   color: var(--tag-preview-text-color);
   background-color: var(--tag-preview-background-color);
-}
-
-a.status-preview:hover, a.status-preview:active, a.status-preview:focus {
-  color: var(--tag-preview-text-color);
 }
 
 .status-deprecated {
@@ -179,15 +169,23 @@ a.status-preview:hover, a.status-preview:active, a.status-preview:focus {
   background-color: var(--tag-deprecated-background-color);
 }
 
-a.status-deprecated:hover, a.status-deprecated:active, a.status-deprecated:focus {
-  color: var(--tag-deprecated-text-color);
-}
-
 .status-experimental {
   color: var(--tag-experimental-text-color);
   background-color: var(--tag-experimental-background-color);
 }
 
-a.status-experimental:hover, a.status-experimental:active, a.status-experimental:focus {
-  color: var(--tag-experimental-text-color);
+.extension-status {
+  display: flex;
+  align-items: center;
+  padding-bottom: 1.25em;
+}
+
+.extension-status-explanation {
+  border-left: 1px solid #aaa;
+  padding: 1.25em;
+  opacity: 0.8;
+}
+
+.extension-status-icon {
+  padding: 1.25em;
 }


insectengine:
For the guides, I'd say we keep it simple text link.
Screenshot 2026-01-20 at 8 45 10 AM

I also again question not having "stable" visible as a status instead of the invisible default. This way people know and it could prevent confusion.

_plugins/asciidoctor-extension.rb
```diff
@@ -163,23 +163,33 @@
...
       status = doc.attr('extension-status')
+      
+      if status.nil? || status.empty?
+        extension_statuses = doc.attr('extensions', '').split(',')
+            .map(&:strip)
+            .reject(&:empty?)
+            .map { |s| s.gsub(/[.:]/, '-') + '-extension-status' }
+            .map { |s| doc.attr(s, '') }
+        status = ['experimental', 'preview', 'deprecated'].find { |s| extension_statuses.include?(s) }
+      end
```
gsmet, line 174: This part is a bit dangerous IMO. IIRC, the `extensions` metadata can be a bit broad as they are used for search.

I wonder if maybe you should only take into account the first one? But then we would have to check the metadata is correct and that the first one is ALWAYS the right one.

gsmet (commented):
I'm all for this but I think we need to discuss this part a bit. See inline.

_plugins/asciidoctor-extension.rb
FroMage: The "upstream" part of the docs (https://github.com/quarkusio/quarkus/pull/52100) will always include a definitive status based on the list of extensions per guide, by selecting the "least favourable" status.
I'm not sure how we can fall into this branch if that's the case.

marko-bekhta: mmm I remember seeing the guide link in the extension metadata yaml:

https://github.com/quarkusio/quarkus/blob/39f7eb8c51ff55216bf50626efa46362490e7971/extensions/hibernate-search-standalone-elasticsearch/runtime/src/main/resources/META-INF/quarkus-extension.yaml#L13

if we go with that ? meaning if we get the status from that extension's yaml and "assign" it to the guide it mentions, instead of just taking the list of extensions specified in the adoc ( https://github.com/quarkusio/quarkus/blob/39f7eb8c51ff55216bf50626efa46362490e7971/docs/src/main/asciidoc/hibernate-search-standalone-elasticsearch.adoc?plain=1#L12)

so after going through all yamls we'd have a map of `<guide,  List<extension>>` ? would that be a more reliable source ?

(just thinking out loud 🫣 🙂 )

marko-bekhta: another idea I had was to say something like: 

> this guide mentions extensions that are in the following statuses : ...

and show them in a "table format" with status, and list of extensions in that status

FroMage: I'm not following the discussion here. What does #52100 do then if not fetch the statuses of all extensions per guide, so it can be used here?

marko-bekhta: right, it fetches the statuses of extensions from their yaml files, and then it looks at the `:extensions:` attribute inside of the adoc file to see which extension statuses to "combine"

https://github.com/quarkusio/quarkusio.github.io/blob/26264d73894994489e4c88082fc6acb59f017312/_guides/aws-lambda.adoc?plain=1#L11

and I was asking if "ignoring" the attribute in the guide (`:extensions:` ) and instead using the guide link in the extension's yaml: 
```
guide: "https://quarkus.io/guides/hibernate-search-standalone-elasticsearch"
```
would be better...

I've no idea, maybe that's the same info just in two different places, maybe not 🤷🏻 🫣 🙂

FroMage: There must be something I don't understand. If https://github.com/quarkusio/quarkus/pull/52100 will correctly set the `:extension-status:` for every guide, why can't we use it here instead of trying to recompute it?

holly-cummins:
Where are we with this? We've just had a complaint about the statuses being out of sync, and I think this might help that (as well as being generally useful and good): https://stackoverflow.com/questions/79980670/quarkus-aws-lambda-http-in-preview-state

marko-bekhta:
> We've just had a complaint about the statuses being out of sync

😭 

> Where are we with this?

Me personally somewhere in the undecided/confused area 🙈 🙂. As to why ?! ... My understanding is that:
- a single guide (adoc file) can have a list of extensions in it
- each extension inside it's yaml file (`quarkus-extension.yaml` in the project's `META-INF/resources`) can have a link to a guide.

Now, my confusion comes from:
* what should we take as a source of truth:
    - guide adoc file metadata
    - `quarkus-extension.yaml` 
    - combine both of the above ^
    - or validate the above and fail Quarkus doc build on mismatch?
* if a guide is referenced from multuple extensions -- how do we want to compute the "final" status that we show to the user on the guide page ?
    - the "worst" one wins
    - show all as pairs extension-status
    - something else 


I'd personally go for "validate and fail" + "show pairs of extension-status" .... but ... 🤷🏻 🙂

holly-cummins:
This PR is at risk from the Roq migration. Most site content I can migrate in an automated way. That means any PRs that miss the train can be patched up using the same scripts. However, the plugins needed to be hand-crafted. That work's already been done, so for where we are, I'll either need to redo the plugin work to incorporate this, or this will need to be ported to java after #2683 merges. Since it's been hanging around for a while, and has some open questions, selfishly, I propose we revisit it post-migration. That does mean we shouldn't do any more coding work before the migration.

marko-bekhta:
I'm ok with adding the extension later in ascidoctorj  🙂,
In the meantime, if we could have a decision on how to derive the status, it would be nice 🙂 🤞🏻

holly-cummins:
My view is we should be as DRY as possible: the extension metadata is the preferred source of truth for status, and ideally, the only source of truth. So I'd say:
- We try quite hard to never include a status in a guide, although this may be hard to stick to, since sometimes a feature will be immature even though an extension is mature 
- Where a guide refers to multiple extensions, we list them all and show the status pairs

But I'm perhaps making comments that are more relevant to the pre-https://github.com/quarkusio/quarkusio.github.io/pull/2508 world.

marko-bekhta:
> since sometimes a feature will be immature even though an extension is mature

so maybe ..... we have an extension status -- we'll show it on the main page where all the guides are listed and then at the top of the guide... but at the same time we allow for a "status badge" at a section title level where a feature described in that section is tagged as experimental (or whatever other status it is in != stable)?

edit: here's an example of how we do it in Search: https://docs.hibernate.org/stable/search/reference/en-US/html_single/#architecture-examples-outbox-polling-elasticsearch

holly-cummins:
Sounds perfect.

_plugins/asciidoctor-extension.rb
```diff
@@ -163,23 +163,33 @@
...
           when 'preview'
-            'This extension\'s backward compatibility and presence in the ecosystem is not guaranteed'
+            '<p>The backward compatibility and presence in the ecosystem of extension(s) discussed here is not guaranteed.</p><p>Specific improvements might require changing configuration or APIs, and plans to become <em>stable</em> are under way. Feedback is welcome on our <a href="https://groups.google.com/d/forum/quarkus-dev">mailing list</a> or as issues in our <a href="https://github.com/quarkusio/quarkus/issues">GitHub issue tracker</a>.</p>'
           when 'stable'
-            'This extension\'s backward compatibility and presence in the ecosystem are taken very seriously'
+            '<p>The backward compatibility and presence in the ecosystem of the extension(s) discussed here are taken very seriously.</p>'
           when 'deprecated'
-            'This extension is likely to be replaced or removed in a future version'
+            '<p>The extension(s) discussed here is likely to be replaced or removed in a future version.</p>'
```
rolfedh, line 185: Wording suggestion — text-level feedback, so it should carry over to whichever implementation lands after the Roq migration:

- Align the explanation text with `extension-status.adoc` (the text readers already see on guides today), with one uniform lead-in pattern: "For _X_ extensions, …". This avoids the "extension(s)" construction, which style guides recommend against, and it stays accurate whether a page discusses one extension or several.
- The include's lead-in says "In _experimental_ mode", but these are *statuses* (that's what the FAQ calls them), and "mode" already means something else in Quarkus (dev mode, native mode).
- Light copy edits along the way: "no guarantee of stability **or** long-term presence" (after a negative like "no", plain "or" is standard, and "long-term" takes a hyphen before a noun); "are not guaranteed" ("backward compatibility and presence" is two things — the `stable` text already uses "are"); "plans are under way for them to become _stable_" (it is the extensions that become stable, not the plans); and the `experimental` period moves inside `</p>` — right now it renders below the paragraph, stuck to the FAQ sentence.

The FAQ entry keeps the older wording for now; I can send a small follow-up to align it.

```suggestion
            '<p>For <em>experimental</em> extensions, early feedback is requested to mature the idea.</p><p>There is no guarantee of stability or long-term presence in the platform until the solution matures. Feedback is welcome on our <a href="https://groups.google.com/d/forum/quarkus-dev">mailing list</a> or as issues in our <a href="https://github.com/quarkusio/quarkus/issues">GitHub issue tracker</a>.</p>'
          when 'preview'
            '<p>For <em>preview</em> extensions, backward compatibility and presence in the ecosystem are not guaranteed.</p><p>Specific improvements might require changing configuration or APIs, and plans are under way for them to become <em>stable</em>. Feedback is welcome on our <a href="https://groups.google.com/d/forum/quarkus-dev">mailing list</a> or as issues in our <a href="https://github.com/quarkusio/quarkus/issues">GitHub issue tracker</a>.</p>'
          when 'stable'
            '<p>For <em>stable</em> extensions, backward compatibility and presence in the ecosystem are taken very seriously.</p>'
          when 'deprecated'
            '<p><em>Deprecated</em> extensions are likely to be replaced or removed in a future version of Quarkus.</p>'
```

_plugins/asciidoctor-extension.rb
```diff
@@ -163,23 +163,33 @@
...
           else
             ""
           end
-        label_html = %(<a class="status-label status-#{status}" title="#{tooltip}" href="#extension-status-note">#{status}</a>)
-
-        label_block = create_pass_block doc, label_html, {}
+        label_html = %(<div class="extension-status-icon"><span class="status-label status-#{status}">#{status}</span></div>)
+        explanation_inner_html = %(<div class="extension-status-explanation">#{explanation_inner_html}For a full list of possible statuses, check our <a href="/faq/#what-are-the-extension-statuses">FAQ entry</a>.</div>)
+        staus_block_html = %(<div class="extension-status">#{label_html}#{explanation_inner_html}</div>)
+        label_block = create_pass_block doc, staus_block_html, {}
```
rolfedh, line 192: Tiny typo: `staus` → `status`.

```suggestion
        status_block_html = %(<div class="extension-status">#{label_html}#{explanation_inner_html}</div>)
        label_block = create_pass_block doc, status_block_html, {}
```

rolfedh (commented):
Thanks @marko-bekhta — I really like the direction here. The old tooltip was easy to miss (its text never appears on touch devices), and deriving the status from extension metadata removes a whole class of silent failures: nobody has to remember the include, and the banner stays correct when an extension is promoted later. The side-by-side layout in your previews looks clean.

Having caught up on the thread: +1 to listing the extensions with their statuses — your table idea from the January inline discussion, and Holly's "list them all and show the status pairs". For guides that cover several extensions, that answers the reader's real question: *which* extension is the non-stable one.

My two inline comments are about the text rather than the code — a typo fix, and a copy edit of the explanation wording, aligned with `extension-status.adoc` — so they should stay relevant for whichever implementation lands after the Roq migration; they are not a request to resume coding now. Glad to help remove the now-redundant `extension-status.adoc` includes once this work lands in whatever form it takes.

PR #2587, opened 2026-03-27
fix: #2580 - Add redirects for /doc and /docs to /guides

lucasoliveira04:
Adds redirects for:

- /doc → /guides
- /docs → /guides

Motivation:
Some LLMs point to /doc as Quarkus documentation location.

### DIFF
--- a/_redirects/doc.md
+++ b/_redirects/doc.md
@@ -1,0 +1,4 @@
+---
+permalink: /doc/
+redirect_to: /guides/
+---
--- a/_redirects/docs.md
+++ b/_redirects/docs.md
@@ -1,0 +1,4 @@
+---
+permalink: /docs/
+redirect_to: /guides/
+---


holly-cummins:
This looks like a good idea to me (thanks!), but we didn't get a preview to confirm things work. I'll leave it to @gsmet, since I don't make redirects very often.

gsmet:
Frankly, I'm not really excited about it. Not sure why LLMs would be confused. But I'm not strongly against it either.

gsmet (commented):
Do you have an example of an LLM pointing to the wrong page?

Also, I'm confused because the files don't seem to follow what we have in the other files such as: https://github.com/quarkusio/quarkusio.github.io/blob/main/_redirects/awards.md?plain=1 .

holly-cummins:
> Frankly, I'm not really excited about it. Not sure why LLMs would be confused. But I'm not strongly against it either.

If we `s/LLM/humans`, it still seems like a good change to me. I'd expect https://quarkus.io/docs to not 404, since it's such a common url pattern and people might guess it works.

gsmet:
People are weird but fine :).

We need the file to be correct though and in line with what we have. I'm not sure they are, especially I don't know where `redirect_to` comes from?

holly-cummins:
> People are weird but fine :).
> 
> We need the file to be correct though and in line with what we have. I'm not sure they are, especially I don't know where `redirect_to` comes from?

Oh, agreed – I wasn't commenting on the actual change contents (which I'm completely unqualified to do), just on the principle that `/docs` is a nice url and we should have one of our own. :)

PR #2588, opened 2026-03-27
Add blog post on the vibe coded langchain4j agentic coder

mariofusco:
/cc @geoand @cescoffier


geoand (commented):
This is certainly interesting, but I really wonder if we should be publishing these experiments here or not.

Given that lots of people have similar experiments running by now, do we plan to publish all of them? And if we do, what will the impact be on people's perception of the quality of the stuff we deliver?

geoand:
All good, but how do we deal with the next blog post that comes around that does something similar in another domain?

mariofusco:
I submitted this article to InfoQ.

PR #2601, opened 2026-04-10
fix: limit public suffix so work on ruby 3.1

maxandersen:
The recent Dependabot upgrade of addressable from 2.8.0 to 2.9.0 (#2597) pulled in public_suffix 7.0.5, which requires Ruby >= 3.2. The
 bretfisher/jekyll-serve Docker image ships Ruby 3.1.7, causing docker compose up to fail with:

 ```
   public_suffix-7.0.5 requires ruby version >= 3.2, which is incompatible with the
   current version, ruby 3.1.7p261
 ```

 This PR pins public_suffix to < 7 in the Gemfile and updates the lockfile to use 6.0.1, which is compatible with Ruby 3.1. 
 
 Alternative is to fork ` bretfisher/jekyll-serve` but thought I would suggest this one first and see if any hard need for public_suffix v7

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -37,5 +37,8 @@ gem "wdm", "~> 0.1.0" if Gem.win_platform?
 
 gem "webrick", "~> 1.8"
 
+# Pin public_suffix to < 7 for Ruby 3.1 compatibility (bretfisher/jekyll-serve image)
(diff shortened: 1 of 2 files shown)


maxandersen:
we shuold just update to use otherw base jekyl...redoing...

PR #2602, opened 2026-04-10
fix: drop outdated jekyll container and use local built instead

maxandersen:
similar to #2601  but instead of limiting gem updates, replace the outdate jekyll container with a local built one.

Tested with docker compose up


cescoffier (approved):

gsmet:
I think we deployed the fixes I had to do for https://github.com/quarkusio/quarkusio.github.io/pull/2495 in all supported branches.

I will have a look though before we merge.

gsmet:
From what I can see some errors but probably not that problematic:

```
Conflict: The following destination is shared by multiple files.                                                                                                                                                           
                      The written file may end up with unexpected contents.                                                                                                                                                                                                                
                      /home/gsmet/git/quarkusio.github.io/_site/author/jbonofre/index.html                                                                                                                                                                                                 
                       - .html                                                                                                                                                                                                                                                             
                       - .html
```

I wasn't able to fix this one quickly, I fixed a few others for which I will create a PR.

gsmet (approved):

PR #2666, opened 2026-05-31
Add page documenting how to migrate from Spring to Quarkus

holly-cummins:
When I present Quarkus, one of the most common questions I get is “how do I migrate from Spring?” I have content on it in my decks, which I've pieced together, but we don’t have content on it on our site. This is a surprising omission, since we have a spring landing page. 

The reason we don’t have information is partly because the story is evolving, but I don’t think we should let this prevent us from sharing *something*. Documenting the current state of play is useful, and then, when we have more, better, solutions, we can iterate on the content. 

We did have an old migrate-to-quarkus page in source, but nothing linked to it, and it wasn't indexed by google. So I've dropped it, to avoid confusion. I've gone for `spring/migrate` as the url, since we might eventually want to have migrations from other starting points. 

Deep links in preview: https://quarkus-website-pr-2666-preview.surge.sh/spring/ and https://quarkus-website-pr-2666-preview.surge.sh/spring/migrate/


_includes/spring-breadcrumb.html
```diff
@@ -0,0 +1,7 @@
+<section class="full-width-version-bg flexfilterbar">
+  <div class="guideflexcontainer">
+    <div class="docslink">
+      <a class="returnlink" href="{{site.baseurl}}/spring"> Back to Using Spring?</a>
```
cescoffier: I'm not sure about the question mark. But maybe it's just me.

_includes/spring-migrate-content.html
```diff
@@ -0,0 +1,106 @@
...
+        <div class="imagetext-container">
+          <div class="imagetext-text">
+            <h5>Quarkus Spring Migration Skill</h5>
+            <p>A Claude Code skill that guides an AI agent through a full Spring Boot to Quarkus migration &mdash; from reading the codebase to committing the result and opening a draft PR. It inspects your project and decides which modules apply, adapting to your chosen strategy: Spring Compatibility or Native Quarkus.</p>
+            <p class="textCTA"><i class="fa fa-chevron-right"></i><a href="https://github.com/quarkusio/quarkus-skills">Quarkus Skills on GitHub</a></p>
+          </div>
+        </div>
+      </div>
+    </div>
+  </div>
```
cescoffier, line 104: No more link to the book?

_includes/spring-migrate-content.html
```diff
@@ -0,0 +1,106 @@
...
+      </div>
+    </div>
+
+    <h3>AI</h3>
+    <div class="grid-wrapper">
+      <div class="width-6-12 width-12-12-m">
+        <div class="imagetext-container">
+          <div class="imagetext-text">
+            <h5>AI-Assisted Migration</h5>
+            <p>Modern AI coding assistants understand both Spring and Quarkus patterns well. They can help translate controllers, services, repositories, configuration files, build files, and tests &mdash; giving you a head start on each file you need to migrate. For a <a href="https://www.the-main-thread.com/p/spring-to-quarkus-migration-guide">hands-on walkthrough</a> of what a full migration looks like step by step, see Markus Eisele's guide on The Main Thread.</p>
```
cescoffier: Found this a bit too generic. I read it and don't know what to do.

_includes/spring-roles-band.html
```diff
@@ -65,7 +65,7 @@ <h4>Boost your Java productivity</h4>
           </div>
           <div class="imagetext-text"><H5>Using Spring now? you don’t have to start over.</h5>
```
cescoffier: Unrelated to this PR, but why an uppercase H here?

cescoffier (approved):
Just a few minor comments. Let's discuss the book stuff.

_includes/spring-breadcrumb.html
holly-cummins: I thought it looked a bit odd, too. It's the literal title from the previous page, but maybe we should be less literal. Removing the question mark makes sense, we're certainly not obliged to be so strict, since it's manually generated (why is it manual?! but that's another topic :) )

When I was proofreading, it almost looked to me like it was suggesting you go back to spring from Quarkus. Clearly we don't want that. So maybe it should be 'Back to Spring landing page'. WDYT?

_includes/spring-migrate-content.html
holly-cummins: My first version had both book links, same as on the landing page, but I took them out, for a few reasons:
- They took a LOT of screen real estate, and since the migration content isn't that long, they visually dominated and made it look like a page about the books. 
- The books don't directly explain how to migrate, more 'why Quarkus is great' (for Quarkus in Action) and map-your-spring-knowledge-to-Quarkus for Eric's. 
- The books are already linked on the Spring landing page 

But maybe the thing to do is to have just a sentence at the bottom, without a huge picture, linking to Eric's book and explaining more specifically why it might help people migrating. WDYT?

_includes/spring-migrate-content.html
holly-cummins: Yeah, I know what you mean. I wanted to have a section saying "just use Claude" since that's basically our advice. And I didn't want to go into sample prompts or that kind of thing, since that felt too detailed, and I didn't want to go into skills or the stuff we're developing, since that's the next section. But I guess it should say it will work without skills but skills make it better. And maybe we can say a simple prompt like "Convert this Spring Boot application to Quarkus" will give good results. I just tried it and unfortunately on the first attempt it tried to convert my spring boot 4 application to Quarkus 4, but I suspect that's just context contamination. :)

_includes/spring-roles-band.html
holly-cummins: Good question. I'll fix while I'm here ...

_includes/spring-breadcrumb.html
cescoffier: Yes, much better

_includes/spring-migrate-content.html
cescoffier: Yes, just a small 'teaser'

_includes/spring-migrate-content.html
cmoulliard: It it time to include the SKILL we created and that users can use as such: https://github.com/quarkusio/skills/blob/main/skills/migrate-spring-to-quarkus/SKILL.md

Long term, we can even promote the new Java acp tool I created as it simplifies the life to use an AI agent (claude, opencode, gemini, ai, etc) and will help users to migrate without breaking the existing project as project is backup-ed when we launch the command "acp -p migrate spring to quarkus" - https://github.com/snowdrop/acp-java-client

holly-cummins: Already done! Your skill is the next section:

image

I've just checked and the link I used is the top level, https://github.com/quarkusio/quarkus-skills, rather than the deep-link blob. I'm not sure which is better. Probably a heading in the readme is best. There's already one there, so I'll do that: https://github.com/quarkusio/skills#migrate-spring-to-quarkus

cescoffier (approved):

PR #2681, opened 2026-06-08
Add blog post analyzing reduced Quarkus performance under rootless Podman

franz1981:
This is the same article as https://redhatperf.github.io/post/hidden-cost-rootless-container-networking/


_data/authors.yaml
```diff
@@ -1,3 +1,9 @@
+fnigro:
+  name: "Francesco Nigro"
+  email: "fnigro@redhat.com"
+  emailhash: "2c6af69cf6a15dbb87f6cb0463ddd986"
+  job_title: "Principal Software Engineer"
```
franz1981: IDK if i should write senior principal or not in IBM @gsmet ^^

_posts/2026-06-08-hidden-cost-rootless-container-networking.adoc
```diff
@@ -0,0 +1,153 @@
...
+== Takeaways
+
+* **A benchmark that saturates an unexpected resource and component is not measuring what you think.** This is what Brendan Gregg calls https://www.brendangregg.com/activebenchmarking.html[active benchmarking]:
++
+[quote, Brendan Gregg]
+____
+_You benchmark A, but actually measure B, and conclude you've measured C._
+____
++
+We benchmarked framework throughput, but Quarkus was saturating pasta's single CPU core — so we were measuring pasta's forwarding capacity, not framework performance. Only by collecting https://man7.org/linux/man-pages/man1/mpstat.1.html[mpstat] and flamegraphs _during_ the run — as https://github.com/quarkusio/spring-quarkus-perf-comparison/issues/62[required by our benchmarking practice] — did we identify which resource and component was saturated. Without that, the 1.19x ratio would have been taken at face value.
```
geoand, line 132: We should add a new line before this because in the render it's too close to the quote above

geoand (commented):
Very cool! 

Is there any chance we can provide some info about `docker` as well?

_data/authors.yaml
gsmet: I would say: do as you prefer, there's no strict rules here. I'm not even sure it can be seen on the website.

gsmet (approved):
I read the content on the other side and it's really good.

Let me know when it's ready to get merged.

franz1981:
@geoand 

> Is there any chance we can provide some info about docker as well?

I actually tried it and is much worse EVEN with network host :/ but TBH i didn't investigated further why is that bad...let me search if I have something

franz1981:
@gsmet ready to go by me!

geoand:
> I actually tried it and is much worse EVEN with network host :/ but TBH i didn't investigated further why is that bad...let me search if I have something

The reason I ask is because Docker is very popular...

franz1981:
@geoand 

PTAL

> The reason I ask is because Docker is very popular...

I know, but tbh the issue was only on podman as it as this "pasta" proxy - while docker has other problems...
I can publish another article to cover specifically docker, but this one is already long IMHO, but the message is clear  to me, regardless: while benchmarking is important to make sure you're measuring what you intend to measure and...Quarkus is so good that can push hard on the limit of the infrastructure, while Spring, nope

geoand:
Yeah, that definitely makes sense

PR #2684, opened 2026-06-08
Handle both Roq and Jekyll in CI

holly-cummins:
If (when?) we convert our site to Roq, it will be a huuuuge PR with the output of a converter. We'll have to run the conversion many times in the process of getting it right. I'm trying to ensure the generated conversion PR doesn't need to include any human-made changes, by making any changes we need to convert to Roq independently. This updates our CI to handle both; we'd probably only build one at a time, but it gives us flexibility. 

The Roq path of this can be seen in https://github.com/quarkusio/quarkusio.github.io/pull/2684; that CI is broken, waiting for a Roq release, but it does correctly attempt to build Roq and not Jekyll.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,7 +20,17 @@ jobs:
           fetch-depth: 5000
           fetch-tags: false
 
+      - name: Check for Jekyll site
+        id: check_jekyll
+        run: |
+          if [ -f "Gemfile" ]; then
+            echo "jekyll_exists=true" >> $GITHUB_OUTPUT
+          else
+            echo "jekyll_exists=false" >> $GITHUB_OUTPUT
+          fi
+
       - name: Set up ruby
+        if: steps.check_jekyll.outputs.jekyll_exists == 'true'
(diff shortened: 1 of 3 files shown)


gsmet (approved):

.github/workflows/build.yml
```diff
@@ -31,18 +41,44 @@ jobs:
...
+            echo "maven_exists=true" >> $GITHUB_OUTPUT
+          else
+            echo "maven_exists=false" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Set up JDK
+        if: steps.check_maven.outputs.maven_exists == 'true'
+        uses: actions/setup-java@be666c2fcd27ec809703dec50e508c2fdc7f6654 # v5.2.0
+        with:
+          java-version: '21'
```
gsmet, line 74: We should probably use Java 25 in these workflows.

holly-cummins: Good point. My first attempt was Java 17, and various tools went 'uh, no'. :) 

I don't know why I only updated to 21, since I was taking the trouble to update, but I'll fix that now(ish).

PR #2690, opened 2026-06-09
Add missing front matter to old Insights transcripts

holly-cummins:
I spotted two old blogs with missing frontmatter. I think Jekyll silently ignores them, but Roq publishes them with today's date. Rather than deleting the files, I patched up the frontmatter, since they're kind of interesting, historically.

### DIFF
--- a/_posts/insights_0_transcription.adoc
+++ b/_posts/insights_0_transcription.adoc
@@ -1,10 +1,18 @@
+---
+layout: post
+title: 'Quarkus Insights #0 transcription: Inception'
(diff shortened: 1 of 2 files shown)


gsmet (approved):

holly-cummins:
Ohh, hang on, I just checked the new content on the live site, and I realised that the missing frontmatter was intentional, and it was an include from https://quarkus.io/blog/insights-1/. 

Rats. Sorry. :(

Will fix it (somehow).

PR #2697, opened 2026-06-10
Add external link checking

### LINKED ISSUE #2110: Add dead link checker to CI
We should not be making our users identify and report dead links for us (for example, most recently #1987 and #2093). Computers can do this job more effectively than people, and unlike people, they don't become disappointed in us when they find dead links. 

We want to catch new dead links when they're introduced, and alert maintainers when existing links stop working. Getting that exact behaviour is non-trivial, but we can get close to it by failing the CI for problems with internal links, and raising a defect if external links stop working. It's possible that a PR would introduce a new external link that never worked at all, but it's more usual that links work when they're created, and then stop working.

The http://hub.quarkiverse.io and http://quarkus.io/extensions pages both use this model, so we should be able to copy their implementations.

holly-cummins:
Resolves #2110 

Draft, because there are still too many dead links. Raising issues for all of them would create too much noise, and instead the job refuses and fails. 

Did you know we have 4,700 pages in the site, which contain 17,973 external links checked?

External links can go AWOL for all sorts of external reasons, so we don't want to block PRs for them. However, we do want to discover them when they happen. The best way to do this is to run a scheduled job and raise defects. This is what we currently do on the extensions site. The length of the job is a bit of a challenge, and false positives caused by bot-blockers is a bigger challenge. Usually I work around that by just having a 'do not check' list, so I've used that pattern here. I've also excluded obvious placeholder links like myserver.com, and links which are in a sentence following "such as", such as "a url such as http://plausible.com". 

The link checker will run in dry run mode for PR changes which directly affect the link checker (like this one), and otherwise it will run as a scheduled job and raise defects. I don't know if the default token will be good enough to raise guide defects in quarkus repo, but we will find out. 

The normal link checker will also check external links, but only for changes which affect a few files, not for infra-level changes. This should help prevent shipping new content with dead links (which happens quite often) without creating too much noise for existing dead links (of which there are many).

image


holly-cummins:
I noticed CI is red for these, but that's related to a dead link or two that came in on the Quarkus side, because we're not doing pre-merge checks on guides. Will fix independently.

PR #2700, opened 2026-06-11
Fix dead podman links

holly-cummins:
Discovered while testing #2697.
Podman no longer have individual endpoints for platform installs, just one big landing page.


holly-cummins:
Oh rats, wrong repo! I forgot this was a guide.

PR #2701, opened 2026-06-11
Improve robustness of dead link checker by looking more closely at potential redirects

holly-cummins:
This finds some links that #2685 misses because it was too trusting when redirects were a possibility, so I've also included a temporary fix for one of our dead links. 

It also fixes some false positives where redirects to the main site would have problems by ensuring we're checking redirects against the locally built site (by rewriting the url, which CI bakes in as quarkus.io).

### DIFF
--- a/_includes/spring-migrate-content.html
+++ b/_includes/spring-migrate-content.html
@@ -21,7 +21,6 @@
       </div>
       <div class="width-6-12 width-12-12-m">
         <ul>
-          <li><a href="{{site.baseurl}}/guides/spring-tx">Spring Transactions</a> &mdash; <code>@Transactional</code></li>
           <li><a href="{{site.baseurl}}/guides/spring-cache">Spring Cache</a> &mdash; <code>@Cacheable</code>, <code>@CacheEvict</code></li>
           <li><a href="{{site.baseurl}}/guides/spring-boot-properties">Spring Boot Properties</a> &mdash; <code>@ConfigurationProperties</code></li>
           <li><a href="{{site.baseurl}}/guides/spring-scheduled">Spring Scheduled</a> &mdash; <code>@Scheduled</code> for periodic tasks</li>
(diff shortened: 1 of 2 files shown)


_includes/spring-migrate-content.html
```diff
@@ -21,7 +21,7 @@ <h3>Spring APIs that work on Quarkus, out of the box</h3>
       </div>
       <div class="width-6-12 width-12-12-m">
         <ul>
-          <li><a href="{{site.baseurl}}/guides/spring-tx">Spring Transactions</a> &mdash; <code>@Transactional</code></li>
+          <li><a href="{{site.baseurl}}/version/main/guides/spring-tx">Spring Transactions</a> &mdash; <code>@Transactional</code></li>
```
gsmet: I'm not really excited about this one. We should probably not mention it for now or we can wait for 3.37 to be released if it includes it.

But pointing to the doc from `main` is not ideal.

holly-cummins: I think it's better than having the dead link, right? And I think having it in the list is important, to indicate that it's coming real soon now. I mean, it doesn't matter all that much because there will be a follow-on PR in two weeks that sorts it out, but I'd like to keep it there to show what's coming.

holly-cummins: I extended this PR to also handle some bad flakiness in the tests, so I think we need to get this in, so I've just removed the problem link. I raised https://github.com/quarkusio/quarkusio.github.io/pull/2716 to put it back next week. Since it's only a week, the exact nature of the solution to the missing content doesn't matter all that much.

holly-cummins:
I've also updated this to fix some egregious false negatives that were breaking most CIs (oops). The issue was that there was a race condition in handling redirects, since they end up all being pointed at quarkus.io by the CI build.

gsmet (approved):

PR #2711, opened 2026-06-12
Include more blog posts in RSS feed

### LINKED ISSUE #2421: Extend https://quarkus.io/feed.xml to cover more months
Right now, the feed only seems to cover a couple of months going back to April. I would suggest to extend this to cover all blog posts on quarkus.io.

holly-cummins:
Resolves https://github.com/quarkusio/quarkusio.github.io/issues/2421 (and also adds tests, since it would be easy to lose the rss feed and not notice).

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -241,7 +241,7 @@ jekyll-archives:
 
 ############################################################
 feed:
-  posts_limit: 50
+  posts_limit: 400
 
 include:
   - ".well-known"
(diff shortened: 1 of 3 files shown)


cescoffier (approved):

_config.yml
```diff
@@ -241,7 +241,7 @@ jekyll-archives:
 
 ############################################################
 feed:
-  posts_limit: 50
+  posts_limit: 400
```
cescoffier, line 244: That's a lot. Did you check the size of the RSS?

cescoffier (approved):

_config.yml
holly-cummins: I didn't, no. :) 

@myfear suggested we include everything, and I increased the amount we put in, but kept it capped. I went to check the size of the file on surge, and it's saying 'project not found,' which is concerning in itself.

PR #2715, opened 2026-06-12
Fix deploy failures caused by incorrect Roq guards

holly-cummins:
As part of my parallel-workstreams-merge-dance, PR #2684 updated build.yml to check for src/main (Roq project) instead of pom.xml, but I missed deploy.yml. It still assumed there were no e2e tests, which was a correct assumption when I wrote the PR, but not when it merged. The deploy problem doesn't show up in PRs, but it does show up in deploys from main. Oooops. 

The test-only pom.xml (for Playwright E2E tests) was triggering the Quarkus Maven build, which fails because there is no Quarkus plugin declared.

I'm also seeing fragility with the link checks that I didn't see on my initial PRs, which might make this CI red. (Why does it work like that? WIP, nice and green in CI, then post-merge, sea of red?!) I've cooked some fixes for the issues, but because there's two problems, neither branch is clean. Still working on piecing that together.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -57,17 +57,17 @@ jobs:
         if: steps.check_jekyll.outputs.jekyll_exists == 'true'
         run: bundle exec jekyll build
 
-      - name: Check for Maven project
-        id: check_maven
+      - name: Check for Roq project
+        id: check_roq
         run: |
-          if [ -f "pom.xml" ]; then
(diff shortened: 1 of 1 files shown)


holly-cummins:
As suspected, this fails with two link checker failures, but they're unrelated to this PR. This PR only changes the deploy step, which isn't exercised at all by PR CIs (hence the original problem).

gsmet (approved):

PR #2723, opened 2026-06-15
CI: add build report to job summary

holly-cummins:
Now that we have tests, it's nice to get a summary of them in the CI status. I used the quarkusio/action-build-reporter to parse the Surefire XML test results and render them.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -198,6 +198,28 @@ jobs:
         env:
           PLAYWRIGHT_SKIP_VALIDATE_HOST_REQUIREMENTS: 1
 
+      - name: Test report
+        if: always()
+        run: |
+          {
+            echo "### Test Results"
+            echo ""
+            echo "| Suite | Tests | Passed | Failed | Skipped | Time |"
+            echo "|-------|------:|-------:|-------:|--------:|-----:|"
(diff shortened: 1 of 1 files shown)


holly-cummins:
Example output: 
image

I wish the actual change was a bit smaller, but of course I didn't want to use a third-party action.

.github/workflows/build.yml
```diff
@@ -198,6 +198,28 @@ jobs:
         env:
           PLAYWRIGHT_SKIP_VALIDATE_HOST_REQUIREMENTS: 1
 
+      - name: Test report
```
gastaldi, line 201: Can we use the https://github.com/quarkusio/action-build-reporter instead?

holly-cummins: I think I tried that and it didn't work very well because I wanted each test and it gave me a overview and the output from the jobs wasn't formatted right. But I wasn't thrilled with the big blob of shell script and I agree it would be nice to do so, so I'll try again to confirm it's not what we need.

gsmet: `action-build-reporter` should work. But we can merge this and improve later given it will make things a bit better already.

gsmet (approved):

PR #2729, opened 2026-06-15
Add smoke tests for the guides page

holly-cummins:
These verify the guides page renders with at least 50 guide entries, has a search/filter input, and contains multiple category sections.

We might need to refactor a bit when we redesign the guides page, but hopefully not too much.

### DIFF
--- a/src/test/java/io/quarkusio/GuidesPageTest.java
+++ b/src/test/java/io/quarkusio/GuidesPageTest.java
@@ -1,0 +1,32 @@
+package io.quarkusio;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
+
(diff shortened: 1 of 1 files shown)


src/test/java/io/quarkusio/GuidesPageTest.java
```diff
@@ -0,0 +1,48 @@
...
+
+import static org.junit.jupiter.api.Assertions.*;
+
+public class GuidesPageTest extends BrowserTest {
+
+    @Test
+    void guidesPageRendersWithSubstantialContent() {
+        page.navigate(baseUrl + "/guides/");
+        int guideCount = page.locator(".docslist h4 a[href*='/guides/']").count();
+        assertTrue(guideCount >= 50,
```
gastaldi, line 13: Is there a reason 50 was chosen?

PR #2735, opened 2026-06-16
Enterprise support - page and footer

insectengine:
This adds the Enterprise support band back to the support page with vendor logos. This also includes an updated footer with support vendor logos.


gsmet:
Do we really need the logos in the footer of all pages?

For me it's too much and it won't scale when we will have 5-10 companies.

We have a Support page, we have a Benefactors page, I think that's good enough?

If we want them on the front page somehow, I would rather have a band with the support possibilities, rather than having the logos on all pages.

insectengine:
> Do we really need the logos in the footer of all pages?
> 
> For me it's too much and it won't scale when we will have 5-10 companies.
> 
> We have a Support page, we have a Benefactors page, I think that's good enough?
> 
> If we want them on the front page somehow, I would rather have a band with the support possibilities, rather than having the logos on all pages.

I've returned to the CF footer as before but added a homepage band with the logos as suggested.

holly-cummins:
What page uses `_includes/supporters-footerband.html`? I can see the html file in the diff, but I can't see other pages using it. And is `homepage-benefactors` replacing the old footer on the homepage? Reading the index.html, I see it as an addition, and nothing is being taken away, but I can't map that to what I'm seeing on the front page of the preview.

insectengine:
@holly-cummins - At this point it won't be used. I didn't want to permanently remove it until we're sure.

holly-cummins:
> @holly-cummins - At this point it won't be used. I didn't want to permanently remove it until we're sure.

Got it. I'd suggest tucking it away somewhere on your local filesystem, not in the PR, just to avoid having reviewers spend time reviewing unused content – and then spending extra time debugging why it's unused. :)

PR #2739, opened 2026-06-18
Use .scss extension for scss files

holly-cummins:
This is part of the general sanitisation prior to the roq move. The brands.css page is scss, so it should have the right extension. I've checked and the file that imports doesn't have an extension in the import, so the move should be ok. 

https://quarkus.io/desktopwallpapers/ seems to be the only page that uses it, so that's the page to check in the preview, although it might not be there because of the site-size-reduction.

*Edit:* https://quarkus-website-pr-2739-preview.surge.sh/desktopwallpapers/ seems to look fine.


insectengine (approved):
That was used for the old /brands page that is now defunct as the content lives on CF.

holly-cummins:
> That was used for the old /brands page that is now defunct as the content lives on CF.

Its classes still do seem to be used in some places. Are you thinking we should switch to different classes there and eliminate the stylesheet?

PR #2741, opened 2026-06-18
Add blog post on C2 compiler thread starvation during warmup

franz1981:
See https://redhatperf.github.io/post/when-the-jit-cant-keep-up/


_posts/2026-06-10-when-the-jit-cant-keep-up.adoc
geoand: Are there any references we can point to about this?

geoand: > Compare this to the CFS charts above

Is there any way we can contrast the two in one chart? Asking because when reading this, I've already forgotten what the previous chart looks like

geoand (commented):
Apart from the two small comments I made, I think this is superb!

_posts/2026-06-10-when-the-jit-cant-keep-up.adoc
franz1981: Screenshot_20260622-085843_Chrome

It doesn't show the related src 😢

geoand: Screenshot from 2026-06-22 10-01-01

franz1981: I should try to diff them introducing a new chart, let's see what I can do...
At worse I can show the spring JVM pls one again as is the one which has the biggest gap, while quarkus as already good (that's our selling point ☝️)

franz1981:
I was tempted to make this article more accessible to a wider audience, but given the type of content and how's presented i.e. an investigation story - dropping data and pointers to a more technical audience will appear as asking the reader for a "leap of faith" which really I don't believe it myself - as an engineer :)

For this reason, I preferred to keep visual some of the behaviours (wait/usr C2 data, tps over time) but keep the technical findings and pointers to better understand them intact.
Users can still go straight to the final "Take aways" if they need to quickly draw conclusions if got bored by the abundance of material :imp:

geoand (approved):

franz1981:
@geoand dp we need another review or we can merge it? 
I have tons of LinkedIn posts where to link it 🤠

geoand:
Let's merge

PR #2744, opened 2026-06-18
Remove $baseurl template preprocessing from SCSS

holly-cummins:
This is one of the first things where I've been unable to come up with a roq alternative for something done in Jekyll. Jekyll uses template processing to inject a value into a stylesheet, and then uses that injected value in a variable used elsewhere in css. The value is $baseurl, a path prefix. It would be useful if quarkus.io was hosted in a sub-path of another site, like commhaus.org/quarkus. But we don't have that requirement, and are unlikely to do so, so I think we can reduce complexity by just ditching the $baseurl. 

If we don't want to do that, the other option is to use a maven step or a `sed` in a CI step to edit the css and put in the blank path prefix for $baseurl, but that seems like a lot of complexity for very little gain. 

This PR:

- Removes the `$baseurl` SCSS variable that was injected via Jekyll Liquid preprocessing (`$baseurl: "{{ site.baseurl }}"`)
- Replaces all `url($baseurl + '/assets/...')` with plain `url('/assets/...')` across 8 SCSS files

Builds on #2740, so should be reviewed and merged after that one. .

We need check these pages on the preview to verify background images render correctly:
- [ ] **Homepage** (`/`) — hero section background image (`bg-home-primary.png`)
- [ ] **Guides index** (`/guides/`) — category icons (tutorials, how-to, concepts, reference, PDF) in both light and dark mode
- [ ] **A specific guide** (e.g. `/guides/getting-started`) — guide type icons in the breadcrumb area
- [ ] **Documentation index** (`/version/main/guides/`) — same category icons + quarkiverse icon on quarkiverse extensions
- [ ] **Footer** (any page) — footer background image (`bg-footer.png`), visible on wide screens (≥1024px)
- [ ] **Spring page** (`/spring/`) — title background image
- [ ] **Events page** (`/events/`) — speaker callout card background
- [ ] **1k Contributors page** — page header background image – only I'm not sure this page has a url anymore

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,12 +26,14 @@ jobs:
           if [ -f "Gemfile" ]; then
             echo "jekyll=true" >> $GITHUB_OUTPUT
             echo "site_dir=_site" >> $GITHUB_OUTPUT
+            echo "statics_dir=." >> $GITHUB_OUTPUT
           else
             echo "jekyll=false" >> $GITHUB_OUTPUT
           fi
           if [ -d "src/main" ]; then
             echo "roq=true" >> $GITHUB_OUTPUT
             echo "site_dir=target/roq" >> $GITHUB_OUTPUT
+            echo "statics_dir=public" >> $GITHUB_OUTPUT
           else
             echo "roq=false" >> $GITHUB_OUTPUT
           fi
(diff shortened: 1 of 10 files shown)


holly-cummins:
I've worked through the URLs which use background images, like https://quarkus-website-pr-2744-preview.surge.sh/events/, and all seem fine.

gsmet (approved):

PR #2750, opened 2026-06-19
Add /releases and /eol pages

holly-cummins:
This is something we've been wanting to do for ages, so that people don't have to go to https://endoflife.date/quarkus-framework to find information we should be providing ourselves.

I've added a releases page showing all Quarkus versions with support status, LTS badges, code.quarkus.io links for supported versions, and linked RHBQ/IBM/HeroDevs support badges. I also added an end-of-life page, modelled on https://node.js/eol, explaining EOL consequences with enterprise support options from HeroDevs.

The releases page is linked off the learn menu (and itself links to the /support page). The /eol page is just linked from the /releases page. 

As a tactical first step the data file is manually generated and maintained, but of course my intention is to automate the process of collecting the data (or at least the skeleton of the data).

### DIFF
--- a/_data/projectfooter.yaml
+++ b/_data/projectfooter.yaml
@@ -20,6 +20,8 @@ links:
         url: /userstories
       - page: Roadmap
         url: https://github.com/orgs/quarkusio/projects/13/views/1
+      - page: Releases
+        url: /releases
       - page: Security&nbsp;policy
         url: /security
       - page: Usage
--- a/_data/releases.yaml
+++ b/_data/releases.yaml
@@ -1,0 +1,446 @@
+policy:
+  lts_cadence: "every 6 months"
+  lts_support_duration: "12 months"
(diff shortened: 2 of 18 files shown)


maxandersen:
This should probably include updating the existing /security page which until now listed the overarching version supported ranges. Just to avoid divergence.

holly-cummins:
> This should probably include updating the existing /security page which until now listed the overarching version supported ranges. Just to avoid divergence.

Ah good point, thanks! I hadn't spotted they were listed there. Will update that page to pull from the same data source. (At the moment it's hardcoded in security.adoc.) However, I think I'll do that in a follow-on PR, to not block this one, since it would be an infra change, not a visible one.

Well, I guess we'd want to link to /eol on the security page. So I'll do that as well.

Sanne (approved):
Looking great! Thanks!

PR #2752, opened 2026-06-19
Add Enterprise and EoL to support

insectengine:
Adds reordered Benefactors band to home page
Adds Enterprise and EoL supports to support page (reordered)
Adds HeroDevs to Benefactors page
Removes footer option.


Sanne:
Rebasing so we can see a comprehensive preview including also Holly's work

PR #2756, opened 2026-06-19
Add redirect tests for noteworthy legacy URLs

holly-cummins:
Verify that /continuum, /community, /publications, /migrate-to-quarkus, and /github all land on their expected targets.

### DIFF
--- a/src/test/java/io/quarkusio/RedirectTest.java
+++ b/src/test/java/io/quarkusio/RedirectTest.java
@@ -1,0 +1,81 @@
+package io.quarkusio;
+
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import java.net.URI;
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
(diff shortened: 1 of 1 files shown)


gsmet (approved):

gsmet:
Note that while I understand the rationale of it. I wonder if we're not going a bit too far and it's not going to be in the way of us updating the website.

But I suppose we can adjust if it causes too much pain in the future.

holly-cummins:
> Note that while I understand the rationale of it. I wonder if we're not going a bit too far and it's not going to be in the way of us updating the website.
> 
> But I suppose we can adjust if it causes too much pain in the future.

Yeah, I've wondered the same thing. All these tests are *super* valuable in the migration because there's so many things to break it's impossible for me to check them all manually. Once we're back to only changing context, they might become overhead. Or, they might catch useful things. I think time will tell which it is. For this one, I thought it might be easy for us to accidentally change something and lose our redirects, but maybe that's only likely in a framework change.

I've *tried* to make the tests so they won't need updating when we update content, but there was a lot of LLM involved, given the size of our site. So I expect we will find some tests are too coupled to content or styles, and when we find those, if it's post-migration, we can loosen (if it's easy) or drop (if we think the test is likely to stay brittle).

holly-cummins:
I guess another category of tests we'd definitely want to keep is ones where we generate the data programatically, like my revised security page, and the releases page. It'd be easy for those to stop working.

PR #2766, opened 2026-06-22
CI: use action-build-reporter for test summary

holly-cummins:
As discussed here: https://github.com/quarkusio/quarkusio.github.io/pull/2723#discussion_r3435652643

Replace the hand-rolled shell script that parsed surefire XML with quarkusio/action-build-reporter, consistent with other Quarkus repos.

To make this work, I had to do a bit of output rearrangement. That was maybe the step I missed before, which made the results so disappointing.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -209,27 +209,18 @@ jobs:
         env:
           PLAYWRIGHT_SKIP_VALIDATE_HOST_REQUIREMENTS: 1
 
-      - name: Test report
+      - name: Prepare build reports
         if: always()
         run: |
-          {
-            echo "### Test Results"
-            echo ""
-            echo "| Suite | Tests | Passed | Failed | Skipped | Time |"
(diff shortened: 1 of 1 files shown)


holly-cummins:
Hmm, no, still not right:

image

Any tips or thoughts, @gsmet?

Maybe it's just that my expectations are different. Well, maybe it's half the problem. The question mark isn't ideal, nor is the warning triangle in the failures column. But what's in main at the moment does this:

image

Seeing the little list of what's covered makes me happy, but there's a good argument that it won't scale (which is of course why quarkus builds only report failures). But for now, since it's in such an unobtrusive place, maybe it's ok? Having the durations reported is also kind of useful, so we know what tests to be selective with, in the future.

gsmet:
Addressed here: https://github.com/quarkusio/quarkusio.github.io/pull/2778 .

gsmet:
Note that you can keep your list if you like, it's not a problem :). But at least you have proper reporting in the PR itself using the same infra we have everywhere.

PR #2772, opened 2026-06-22
Add checks for raw html and unrendered markdown on specific pages

holly-cummins:
I don't want to test the whole site because that's (a) slow and (b) makes life a nightmare on pages about raw html and raw markdown, but we don't want raw html in important places like the frontpage.

### DIFF
--- a/src/test/java/io/quarkusio/AuthorPageTest.java
+++ b/src/test/java/io/quarkusio/AuthorPageTest.java
@@ -2,6 +2,8 @@ package io.quarkusio;
 
 import org.junit.jupiter.api.Test;
 
+import static io.quarkusio.UnrenderedMarkupDetector.assertDoesNotContainRawHtml;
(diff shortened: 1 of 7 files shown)


gsmet (approved):
Not sure this has ever been a problem but I suppose it can't hurt if we don't have too many false negatives.

holly-cummins:
> Not sure this has ever been a problem but I suppose it can't hurt if we don't have too many false negatives.

Yeah, I wondered about that for some of these. It has not been a problem in the past, but I've made it a problem! :) 

Running tests for everyone for ever for problems that only affect me isn't great, but it also seems like a sensible thing to protect against because it *could* happen with any site generation framework if people get escapes wrong. 

Once the dust settles, like in a six months or so, we can maybe review and see what tests we think are addding value and which are taking too much time or failing incorrectly.

PR #2781, opened 2026-06-23
Drop search attribute which does not exist in config and is not used

holly-cummins:
The Roq conversion tripped over this – I can hack round it, but we may as well tidy up. 

Part of our search page references `search.url`, a property that is not defined in `_config.yml`. Claude says `search.url` was the original config key, then later `search.host` replaced it. But the `data-search-api-server attribute` on the div was never updated to use `host` instead of `url`. The `data-search-api-server` is likely unused since the web component reads from its own server attribute.

### DIFF
--- a/_includes/index-guides.html
+++ b/_includes/index-guides.html
@@ -1,12 +1,7 @@
 {% assign docversion = include.docversion %}
 {% assign data_source = include.source %}
 
-<div id="guides-app"
-     data-quarkus-version="{{ docversion }}"
-     data-language="{{ site.language }}"
-     data-initial-timeout="{{ site.search.initial-timeout }}"
-     data-more-timeout="{{ site.search.more-timeout }}"
-     data-min-chars="{{ site.search.min-chars }}">
+<div id="guides-app">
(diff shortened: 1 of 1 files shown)


_includes/index-guides.html
marko-bekhta:
```suggestion
<div id="guides-app">
```
maybe let's remove all of them ^ 🫣 🙈 
I think these data attributes were from the initial app (was it vue ? 🤔 ) that was later replaced with lit components, but the data attributes were never removed ?

holly-cummins: Ok, I'll try and we can see what happens 🫣 🫣

marko-bekhta: 🤞🏻 🫣 
I tried locally and didn't notice any errors in the logs/console... but yes, please double-check 🙇🏻 🙂 thanks!

marko-bekhta (approved):
🤞🏻 🙂 
thanks!

holly-cummins:
Search seems to work in preview, so merging.

PR #2782, opened 2026-06-23
Make version table in security.adoc use data from _data

### LINKED ISSUE #2776: Update security page to link /eol page and single source of truth for release data
See https://github.com/quarkusio/quarkusio.github.io/pull/2750#issuecomment-4773613761

holly-cummins:
Resolves https://github.com/quarkusio/quarkusio.github.io/issues/2776

Manually maintaining this table is boring, and unnecessary now that we have a single source of truth about lts versions. 

Annoyingly, we can't do the template interpolation into adoc, so I had to make a franken-page which kept some adoc parts (because adoc is nicer to edit), but used html for the table. 

This should produce a /security which is exactly identical to the current one, with the exception of the link to /eol. 

Before:
https://quarkus.io/security/
image

After:
https://quarkus-website-pr-2782-preview.surge.sh/security
image

### DIFF
--- a/_includes/security-supported-versions.html
+++ b/_includes/security-supported-versions.html
@@ -1,0 +1,47 @@
+{%- assign today = 'now' | date: '%Y-%m-%d' -%}
+{%- assign latest_version = nil -%}
+{%- for release in site.data.releases.releases -%}
+  {%- unless release.upcoming -%}
+    {%- unless latest_version -%}
+      {%- assign latest_version = release.version -%}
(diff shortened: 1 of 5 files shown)


gsmet (approved):
Oh nice!

holly-cummins:
I noticed that there's also versions.yml which overlaps slightly with my new releases.yaml. Having the pair of them is confusing, so I'll try and consolidate if I can – that doesn't affect this work item, but mentioning it for awareness for @gsmet. The overlap is only slight, which makes it harder to figure out the right outcome.

PR #2784, opened 2026-06-24
Add Quarkus Data Hibernate getting started tutorial

lucamolteni:
## Summary

Step-by-step guide covering Quarkus Data Hibernate repositories: entity mapping, repository pattern, managed vs stateless sessions, native SQL queries, and Hibernate Reactive.


lucamolteni:
@FroMage @yrodiere take a look at it when you can

_guides/quarkus-data-hibernate-getting-started.adoc
```diff
@@ -0,0 +1,811 @@
...
+
+You're working on a physical bookstore website, in which you want users to be able to browse books with titles.
+In this tutorial, we won't deal with the frontend and the user interface, we just want to connect to the database.
+
+We'll start this tutorial by creating a new empty Quarkus application. The easiest way to do so is to have the Quarkus CLI tool available:
+
+[source,bash]
+----
+quarkus create app --name="Quarkus Data Tutorial" quarkus-data-tutorial --platform-bom=io.quarkus.platform:quarkus-bom:3.37.0.CR1
+----
```
yrodiere: There are asciidoc "partials" that you can include for this, and will display a nice UI where you can chose between the CLI, Maven or Gradle.


E.g. https://quarkus.io/guides/hibernate-orm#setting-up-and-configuring-hibernate-orm

Image

Or https://quarkus.io/guides/getting-started#bootstrapping-the-project

Image

Maybe use something similar here?

_guides/quarkus-data-hibernate-getting-started.adoc
```diff
@@ -0,0 +1,811 @@
...
+
+Let's start by importing the extensions:
+
+[source,bash]
+----
+cd quarkus-data-tutorial
+
+quarkus extension add quarkus-data-hibernate
+quarkus extension add quarkus-jdbc-postgresql
```
yrodiere: Some comments, take what you like:

1. Same as above, you might want to present CLI/Maven/Gradle
2. You can add extensions directly in the first CLI call (`quarkus create app`), no need for a separate call
3. If you really want to split this, you can probably use a single CLI call to add both extensions, which would likely be a bit faster. Not sure if extensions are comma-separated or space-separated though.

_guides/quarkus-data-hibernate-getting-started.adoc
```diff
@@ -0,0 +1,811 @@
...
+----
+
+The first dependency, `quarkus-data-hibernate`, is the Quarkus extension that provides everything you need for database access.
+The second dependency, `quarkus-jdbc-postgresql`, is the JDBC driver that tells Quarkus how to talk to a PostgreSQL database. If you're using a different database (MySQL, MariaDB, H2, etc.), pick the corresponding `quarkus-jdbc-*` driver instead. Quarkus Data Hibernate cannot work without a JDBC driver.
```
yrodiere: A link to the relevant section of the datasource guide that lists available extensions would likely make sense here?

_guides/quarkus-data-hibernate-getting-started.adoc
```diff
@@ -0,0 +1,811 @@
...
+
+That's all you need to start querying your database.
+
+Quarkus Data Hibernate uses the Hibernate annotation processor, a key component that reads your code at compile time, validates queries, and generates implementation classes under `target/generated-sources/annotations/`.
+These generated classes are compiled together with your code and wired up automatically by Quarkus. They are also simple to read, in case you want to understand how it works internally.
```
yrodiere: Not sure this path holds for Gradle.

```suggestion
Quarkus Data Hibernate uses the Hibernate annotation processor, a key component that reads your repository interfaces at compile time, validates queries, and generates implementation classes.
These generated classes are compiled together with your code and wired up automatically by Quarkus. They are also simple to read, in case you want to understand how it works internally.
```

_guides/quarkus-data-hibernate-getting-started.adoc
```diff
@@ -0,0 +1,811 @@
...
+        <annotationProcessorPathsUseDepMgmt>true</annotationProcessorPathsUseDepMgmt>
+        <annotationProcessorPaths>
+            <path>
+                <groupId>org.hibernate.orm</groupId>
+                <artifactId>hibernate-processor</artifactId>
+            </path>
+        </annotationProcessorPaths>
+    </configuration>
+</plugin>
+----
```
yrodiere: Needs Gradle example?

_guides/quarkus-data-hibernate-getting-started.adoc
```diff
@@ -0,0 +1,811 @@
...
+
+The first dependency, `quarkus-data-hibernate`, is the Quarkus extension that provides everything you need for database access.
+The second dependency, `quarkus-jdbc-postgresql`, is the JDBC driver that tells Quarkus how to talk to a PostgreSQL database. If you're using a different database (MySQL, MariaDB, H2, etc.), pick the corresponding `quarkus-jdbc-*` driver instead. Quarkus Data Hibernate cannot work without a JDBC driver.
+
+That's all you need to start querying your database.
+
+Quarkus Data Hibernate uses the Hibernate annotation processor, a key component that reads your code at compile time, validates queries, and generates implementation classes under `target/generated-sources/annotations/`.
+These generated classes are compiled together with your code and wired up automatically by Quarkus. They are also simple to read, in case you want to understand how it works internally.
+
+To enable it, find the `maven-compiler-plugin` in your `pom.xml` (it was generated by `quarkus create app`) and add the annotation processor configuration to it:
```
yrodiere: Also @FroMage do you think `quarkus ext add` could add the processor automatically? I think you looked into this when working on your PR to run processing even if this configuration is absent, no?

_guides/quarkus-data-hibernate-getting-started.adoc
```diff
@@ -0,0 +1,811 @@
...
+----
+
+Now run Quarkus continuous testing in a separate tab of your terminal:
+
+[source,bash]
+----
+quarkus test
+----
+
+You don't need to restart Quarkus testing, it will re-run tests continuously after every change.
```
yrodiere, line 60: Is the use of `quarkus test` instead of `quarkus dev` an explicit choice? I've always felt that dev mode, with the swagger UI, was particularly useful to try things out, but it's true that continuous testing is an interesting approach, too.

_guides/quarkus-data-hibernate-getting-started.adoc
```diff
@@ -0,0 +1,811 @@
...
+
+Quarkus Data Hibernate can handle the database schema for you.
+Since our `import.sql` already handles the full schema creation, let's tell Hibernate not to manage it and tell Dev Services to run our script when the database container starts.
+Add this to `src/main/resources/application.properties`:
+
+[source,properties]
+----
+quarkus.hibernate-orm.schema-management.strategy=none
+quarkus.datasource.devservices.init-script-path=import.sql
+----
```
yrodiere: As discussed, I think this can be a bit overwhelming for a getting started tutorial that's supposed to introduce the basics, especially in the first steps. I think there are things to improve here -- either in your tutorial or in Quarkus -- because it's clearly not smooth.

First, the schema could be generated by Hibernate, which would be a much better experience for development. But then when you start your tutorial, you don't have entities, so that forces you do all this -- including overriding defaults with `quarkus.hibernate-orm.schema-management.strategy=none`, even though the default of `drop-and-create` is there precisely to help!
One solution to this would be to split this into two separate guides, BTW: one for pure SQL and one for mapping. It could make sense because the audience is not the same: people doing entity mapping will use SQL only in very specific edge cases, while people using SQL all over the place will not want to even hear about Hibernate. Also, people using SQL all over the place will be much more comfortable with a big DDL script at the start of the tutorial.

Second, this is mixing schema and data population, which in actual applications (when going further) would be separate concerns. It'd be nice if we clearly separated them here. Unfortunately until we solve #21866 , the only solution to separate data population would be to move it to the test directly.
Though if things stay simple, creating the data in the test is an idea too: you could just have a `@BeforeAll` method that calls the repository to add instances?

_guides/quarkus-data-hibernate-getting-started.adoc
```diff
@@ -0,0 +1,811 @@
...
+
+TIP: JDBC connections are called "Data Sources" in Quarkus. In dev mode, if you don't configure a datasource, Quarkus will automatically start a PostgreSQL container for you via Dev Services. This means you can run `quarkus dev` and start coding without installing or configuring a database. See the xref:datasource.adoc[Datasource guide] for more details.
```
yrodiere: This probably deserves to be explained before you talk about `quarkus test`, because people will wonder: wait, how could tests work, I don't have a database yet -- damn, I need to install one, I'll do that first...

_guides/quarkus-data-hibernate-getting-started.adoc
```diff
@@ -0,0 +1,811 @@
...
+    List<BookRecord> findByTitle(String title); // <5>
+
+    @SQL("SELECT id, title, isbn, price FROM Book WHERE proce < :maxPrice")
+    List<BookRecord> cheaperThan(BigDecimal maxPrice);
+}
+----
+<1> A repository interface that groups all your query methods for books. The annotation processor generates the implementation for you.
+<2> A Java record that maps to the columns returned by the SQL query. Each field name must match a column name in the result set.
+<3> `@SQL` sends raw SQL to the database. The `:title` syntax is a named parameter, it avoids SQL injection attacks.
+<4> `@SQL` is a Hibernate-specific annotation. Jakarta Persistence 4.0 will introduce a standard https://github.com/jakartaee/persistence/pull/992[`@NativeQuery`] equivalent.
```
yrodiere: No need to explain this in a tutorial IMO. This comment is really for us.

```suggestion
// TODO Replace @SQL with Jakarta Persistence 4.0's standard https://github.com/jakartaee/persistence/pull/992[`@NativeQuery`] equivalent.
```

_guides/quarkus-data-hibernate-getting-started.adoc
```diff
@@ -0,0 +1,811 @@
...
+[source,java]
+----
+    @SQL("SELECT id, title, isbn, price FROM Book WHERE price < :maxPrice") // <1>
+    List<BookRecord> cheaperThan(BigDecimal maxPrice);
+----
+<1> Change `proce` to `price` here
+
+and then all the tests will now pass.
+
+Writing SQL by hand makes sense for very simple scenarios, or when you need database-specific features such as window functions, recursive functions and similar advanced patterns.
```
yrodiere: I think that's a fine point to make in a talk, but:

1. It's odd to give that kind of lecture in a getting started tutorial. People are here to get started, fast, and they already want to use Quarkus Data. They do not need to be convinced they should use it. IMO this would find a better home in whatever "main" page we decide to use for the Quarkus Data documentation, because *there* we need to explain to people what to use, and why. That page cannot be a tutorial IMO, it needs to be short and to the point.
2. The last statement is antagonizing a whole class of advanced users who prefer to write SQL by hand, just because they know SQL and want control over it -- essentially, people who like Spring Data JDBC. I don't think it's a good idea to tell them they are wrong, because at the core this is a matter of taste, and you will never convince them.

_guides/quarkus-data-hibernate-getting-started.adoc
```diff
@@ -0,0 +1,811 @@
...
+[source]
+----
+error: no matching field named 'tilte' in entity class 'org.acme.Book'
+----
+
+Get the parameter back to the correct name so that we can proceed further.
+
+== Creating new entities
+
+So far we've been reading data, but our bookstore is crowdsourced: users can add books they want to share.
```
yrodiere: Not sure you need to build a story for this. We just need to persist data, period -- you don't have to justify yourself :)

_guides/quarkus-data-hibernate-getting-started.adoc
```diff
@@ -0,0 +1,811 @@
...
+We need to persist new books to the database.
+
+Let's update `Book` to implement `PanacheEntity.Stateless`, which gives us lifecycle operations directly on the entity:
```
yrodiere: I would do this from the start, because this step unnecessarily complicates your tutorial. We're already straying quite far from the "straight to the point" approach in [the Spring Data JPA getting started guide](https://docs.spring.io/spring-data/jpa/reference/jpa/getting-started.html).

You can just omit explaining the lifecycle operations until you get to this section, and explain them right when you show how to implement the write operation.

_guides/quarkus-data-hibernate-getting-started.adoc
```diff
@@ -0,0 +1,811 @@
...
+
+[NOTE]
+====
+If you're an experienced Hibernate user, you'll find this part very different from what you're used to.
+When you call `insert()`, `update()`, or `delete()`, an operation is instantly triggered against the database, without calling `.flush()` explicitly.
+Managed entities are nowhere to be seen, and neither is the idea of detached and persisted objects.
+This is intentional: it's a simpler model that uses `StatelessSession` under the covers.
+In this mode every change to an object requires an explicit save operation.
+If you're not interested in this style and want the full power of managed entities, skip ahead to <<managed-session>>.
+====
```
yrodiere, line 186: This complicates the tutorial. Leave it for the section about managed sessions?

_guides/quarkus-data-hibernate-getting-started.adoc
```diff
@@ -0,0 +1,811 @@
...
+=== About entity identifiers
+
+`WithId.AutoLong` is more powerful than Hibernate's traditional `@Id @GeneratedValue` pair: there are variants for different ID types out of the box:
+
+- `WithId.AutoLong`, a generated `Long` id
+- `WithId.AutoUUID`, a generated `UUID` id
+- `WithId.AutoString`, a generated `String` id
+
+Of course you can always use `@Id` and `@GeneratedValue` directly if you prefer, or if you need a custom ID strategy.
+See the xref:quarkus-data-hibernate.adoc#_about_entity_identifiers[reference guide] for the full list of options.
```
yrodiere, line 194: IMO this should just be a link to the reference documentation for Quarkus Data in your code example about writing an entity. E.g. on `extends WithId.AutoLong`, you say `this defines IDs as auto-generated longs; other types of IDs have convenient shortcuts or advanced customization options, see xref:quarkus-data-reference.adoc#ids[this section of the reference]`.

Of course, that requires that we have a Quarkus Data reference in the first place :)

_guides/quarkus-data-hibernate-getting-started.adoc
```diff
@@ -0,0 +1,811 @@
...
+When you call `insert()`, `update()`, or `delete()`, an operation is instantly triggered against the database, without calling `.flush()` explicitly.
+Managed entities are nowhere to be seen, and neither is the idea of detached and persisted objects.
+This is intentional: it's a simpler model that uses `StatelessSession` under the covers.
+In this mode every change to an object requires an explicit save operation.
+If you're not interested in this style and want the full power of managed entities, skip ahead to <<managed-session>>.
+====
+
+=== About entity identifiers
+
+`WithId.AutoLong` is more powerful than Hibernate's traditional `@Id @GeneratedValue` pair: there are variants for different ID types out of the box:
```
yrodiere: It is not more powerful, quite obviously, since it's based on Hibernate ORM :)

It's convenient syntactical sugar.

_guides/quarkus-data-hibernate-getting-started.adoc
```diff
@@ -0,0 +1,811 @@
...
+        // Other business logic that modifies books i.e. applying loyalty points, rounding prices ...
+        // Since we want to update them only once, we collect them to avoid multiple updates within the loops
+        for (Book book : modified) {
+            bookRepo.update(book); // <4>
+        }
+    }
+}
+----
+<1> We collect the books we modify so we can update them all at the end, after all business logic has run. Updating inside the body will result in multiple SQL update statements.
+<2> In a stateless session, lazy associations are not fetched automatically. We must load the publisher explicitly. A join fetch query in the Book repository could be used instead of a separate fetch, but for the scope of the tutorial we're not introducing join fetches.
```
yrodiere: IMO you're getting too far here. You're addressing beginners -- if you think join fetches are too complicated to explain here, you shouldn't mention them. Also, beginners might not even know what a lazy association is, so we'd need to explain that.

Perhaps more importantly, this is rather bad code. In fact, it's textbook N+1 select.

I think it would make more sense to explain how one would *properly* implement this with a stateless session. If you keep a ToOne association here, it would need to be eager, and if you replace it with a ToMany association, it would make sense to explain join fetches, because they would be a core feature for any application using stateless session.

_guides/quarkus-data-hibernate-getting-started.adoc
```diff
@@ -0,0 +1,811 @@
...
+
+        @Query("where price < :maxPrice order by title")
+        List<Book> cheaperThan(BigDecimal maxPrice);
+
+        @Query("where publisher.country = :country")
+        List<Book> findByPublisherCountry(String country);
+    }
+}
+----
+<1> We use `FetchType.LAZY` so that loading a book doesn't automatically join and fetch its publisher.
```
yrodiere: This is kind of odd. Lazy to-ones are not that frequent in my experience. Neither is explicit configuration of laziness, btw -- defaults are generally fine, and probably should be in such a getting started tutorial.

You could find some other example with a to-many associations? That way you don't even need to mention laziness here, and can mention it further down.

_guides/quarkus-data-hibernate-getting-started.adoc
```diff
@@ -0,0 +1,811 @@
...
+        }
+    } // <3>
+}
+----
+<1> Accessing `book.publisher` triggers a lazy fetch: Hibernate loads the publisher from the database automatically with a query.
+<2> We just set the field. No need to call `update()`.
+<3> When the `@Transactional` method returns, Hibernate compares every managed entity against its original state and generates the correct SQL. Books whose publisher didn't allow discounts were never changed, so Hibernate skips them entirely.
+
+The test is exactly the same: the behavior hasn't changed, just the implementation got simpler.
+It's up to you to decide whether to use a Managed Entity or a Stateless one. The Managed Repository is more expressive, but reasoning about it is more complex and might require knowing some details about how Hibernate handles objects. If you're an experienced Hibernate developer, you probably already know the discussions about n+1 queries, and you expect that changing an object's field doesn't immediately hit the database. Some users prefer a more explicit way to control the database. Either way is fine, and Quarkus Data Hibernate makes it easy to use both and to switch without committing to one style or the other.
```
yrodiere:
```suggestion
The test is exactly the same: the behavior hasn't changed, the implementation only got simpler.

It's up to you to decide whether to use a Managed Entity or a Stateless one. The Managed Repository is more expressive, but reasoning about it is more complex and might require knowing some details about how Hibernate handles objects. If you're an experienced Hibernate developer, you probably already know the discussions about n+1 queries, and you expect that changing an object's field doesn't immediately hit the database. Some users prefer a more explicit way to control the database. Either way is fine, and Quarkus Data Hibernate makes it easy to use both and to switch without committing to one style or the other.
```

_guides/quarkus-data-hibernate-getting-started.adoc
```diff
@@ -0,0 +1,811 @@
...
+    void shouldReturnEmptyListForUnknownTitleReactively(UniAsserter asserter) {
+        asserter.assertThat(
+            () -> reactiveBookRepo.findByTitle("Unknown Book"),
+            books -> assertTrue(books.isEmpty())
+        );
+    }
+}
+----
+<1> `@RunOnVertxContext` ensures the test runs on the Vert.x event loop, which is required for reactive database access.
+<2> `@TestReactiveTransaction` opens a reactive session and transaction for the test, and rolls back at the end.
```
yrodiere, line 584: Not about this tutorial per se, but... this is awfully complicated. In an ideal world we'd require only `@TestTransaction` and `@RunOnVertxContext` would be inferred from `UniAsserter`. No?

cc @FroMage

yrodiere (commented):
Thanks, here are a few comments.

The main thing, IMO, is that we probably need to rethink the "start with SQL and go from there" approach for a tutorial. Yes, I know that you had presented that a long time ago, but I don't think it works as well as we had hoped *specifically for this tutorial*, so it's worth reconsidering.

I believe we're trying to conflate three things into this tutorial, and that's too much and should be split:

1. An entry point for Quarkus Data documentation, telling you what you can do and what to chose
2. A getting started guide for entity mapping and repositories with Quarkus Data
3. A getting started guide for pure-SQL repositories with Quarkus data

_guides/quarkus-data-hibernate-getting-started.adoc
```diff
@@ -0,0 +1,811 @@
...
+== Introduction
+
+You're working on a physical bookstore website, in which you want users to be able to browse books with titles.
+In this tutorial, we won't deal with the frontend and the user interface, we just want to connect to the database.
+
+We'll start this tutorial by creating a new empty Quarkus application. The easiest way to do so is to have the Quarkus CLI tool available:
+
+[source,bash]
+----
+quarkus create app --name="Quarkus Data Tutorial" quarkus-data-tutorial --platform-bom=io.quarkus.platform:quarkus-bom:3.37.0.CR1
```
FroMage: I don't think we should hardcode versions there, it will not age well.

_guides/quarkus-data-hibernate-getting-started.adoc
FroMage: I don't know. Perhaps here it's too early, but somewhere, it would be useful to tell people where those files are generated.

_guides/quarkus-data-hibernate-getting-started.adoc
FroMage: I don't think there's machinery for that, not unless it includes the quickstarts code.

Perhaps @aloubyansky or @ia3andy know?

FroMage: It would be nice.

_guides/quarkus-data-hibernate-getting-started.adoc
FroMage: Yes, agreed. This is scary as hell, and only required for freaks like @maxandersen who are allergic to `@Entity`, let's split this off into a file for Max please :)

_guides/quarkus-data-hibernate-getting-started.adoc
```diff
@@ -0,0 +1,811 @@
...
+import org.hibernate.annotations.processing.SQL;
+
+import java.math.BigDecimal;
+import java.util.List;
+
+public interface BookQueries { // <1>
+
+    record BookRecord(Long id, String title, String isbn, BigDecimal price) {} // <2>
+
+    @SQL("SELECT id, title, isbn, price FROM Book WHERE title = :title") // <3> <4>
```
FroMage: I know this is SQL and we're not trying to improve on it, but if people are actually using this a lot, we could… in Jakarta Data, and simplify this to `FROM Book WHERE title = :title` because that first select part can totally be inferred from the return type, just like it does for `@Query`.

WDYT @gavinking ? Should I open a JD issue for this for the future?

_guides/quarkus-data-hibernate-getting-started.adoc
FroMage: Well it is a warning to users that this will change.

_guides/quarkus-data-hibernate-getting-started.adoc
```diff
@@ -0,0 +1,811 @@
...
+
+    @SQL("SELECT id, title, isbn, price FROM Book WHERE proce < :maxPrice")
+    List<BookRecord> cheaperThan(BigDecimal maxPrice);
+}
+----
+<1> A repository interface that groups all your query methods for books. The annotation processor generates the implementation for you.
+<2> A Java record that maps to the columns returned by the SQL query. Each field name must match a column name in the result set.
+<3> `@SQL` sends raw SQL to the database. The `:title` syntax is a named parameter, it avoids SQL injection attacks.
+<4> `@SQL` is a Hibernate-specific annotation. Jakarta Persistence 4.0 will introduce a standard https://github.com/jakartaee/persistence/pull/992[`@NativeQuery`] equivalent.
+<5> The method signature defines the parameters. The annotation processor generates the implementation.
```
FroMage:
```suggestion
<5> The method signature defines the parameters. 
```
The last part was already said at `<1>`

_guides/quarkus-data-hibernate-getting-started.adoc
```diff
@@ -0,0 +1,811 @@
...
+import java.util.List;
+
+public interface BookQueries { // <1>
+
+    record BookRecord(Long id, String title, String isbn, BigDecimal price) {} // <2>
+
+    @SQL("SELECT id, title, isbn, price FROM Book WHERE title = :title") // <3> <4>
+    List<BookRecord> findByTitle(String title); // <5>
+
+    @SQL("SELECT id, title, isbn, price FROM Book WHERE proce < :maxPrice")
```
FroMage: Might want to add a footnote here because that I did not think that typo was intentional on my first read.

_guides/quarkus-data-hibernate-getting-started.adoc
FroMage: Well, there's people who choose this for taste, we can't save those people. 

But most people will pick SQL only for really good reasons, and we should open this guide with those reasons: why should they use SQL? And equally importantly: when should they not use SQL? And point them to the other tutorial in case they got lost here for no good reason.

And a similar link back from the other tutoral in case they _need_ (or want) SQL.

_guides/quarkus-data-hibernate-getting-started.adoc
```diff
@@ -0,0 +1,811 @@
...
+import io.quarkus.hibernate.panache.PanacheRepository;
+import io.quarkus.hibernate.panache.WithId;
+import jakarta.data.repository.Find;
+import jakarta.data.repository.Query;
+import jakarta.persistence.Entity;
+
+import java.util.List;
+
+@Entity // <1>
+public class Book extends WithId.AutoLong { // <2>
```
FroMage:
```suggestion
public class Book extends ManagedEntity { // <2>
```
This is mixing the future packages with the past type names

_guides/quarkus-data-hibernate-getting-started.adoc
```diff
@@ -0,0 +1,811 @@
...
+    public String title; // <3>
+    public String isbn;
+    public BigDecimal price;
+
+    public interface Repo extends PanacheRepository<Book> { // <4>
```
FroMage:
```suggestion
    public interface Repo extends ManagedRepository<Book> { // <4>
```

_guides/quarkus-data-hibernate-getting-started.adoc
```diff
@@ -0,0 +1,811 @@
...
+        List<Book> cheaperThan(BigDecimal maxPrice); // <6>
+    }
+}
+----
+<1> Tells Hibernate that this class maps to a database table. Hibernate will use the class name as the table name and the field names as column names.
+<2> Gives the entity a generated `Long` id field automatically.
+<3> These public fields map to columns in the `Book` table. Hibernate reads and writes these fields directly.
+<4> Defines a repository: a place to put all your query methods for `Book`. By nesting it inside the entity, the queries and the data they operate on live together. You can name this interface whatever you want: `Queries`, `Repo`, `BookQueries`, etc.
+<5> `@Find` tells Quarkus Data Hibernate to generate a query based on the method's parameter names. `findByTitle(String title)` becomes a query that matches the `title` field on `Book`. Parameters are inferred from the method signature.
+<6> `@Query` lets you write an HQL fragment when you need more control. It avoids the need to write `select b from Book`. The entity type is inferred from the return type `List<Book>`.
```
FroMage:
```suggestion
<6> `@Query` lets you write an HQL fragment when you need more control. It avoids the need to write what you select. The entity type is inferred from the return type `List<Book>`.
```

Otherwise, it would be `from Book` or `select b from Book b`

_guides/quarkus-data-hibernate-getting-started.adoc
```diff
@@ -0,0 +1,811 @@
...
+----
+
+You should see a compilation error like:
+
+[source]
+----
+error: no matching field named 'tilte' in entity class 'org.acme.Book'
+----
+
+Get the parameter back to the correct name so that we can proceed further.
```
FroMage, line 119: This is the hardest I've worked following a tutorial 😓

_guides/quarkus-data-hibernate-getting-started.adoc
```diff
@@ -0,0 +1,811 @@
...
+
+So far we've been reading data, but our bookstore is crowdsourced: users can add books they want to share.
+We need to persist new books to the database.
+
+Let's update `Book` to implement `PanacheEntity.Stateless`, which gives us lifecycle operations directly on the entity:
+
+[source,java]
+----
+@Entity
+public class Book extends WithId.AutoLong implements PanacheEntity.Stateless { // <1>
```
FroMage:
```suggestion
public class Book extends RecordEntity { // <1>
```

_guides/quarkus-data-hibernate-getting-started.adoc
```diff
@@ -0,0 +1,811 @@
...
+    public String title;
+    public String isbn;
+    public BigDecimal price;
+
+    public interface Repo extends PanacheRepository.Stateless<Book, Long> { // <2>
```
FroMage:
```suggestion
    public interface Repo extends RecordRepository<Book> { // <2>
```

_guides/quarkus-data-hibernate-getting-started.adoc
```diff
@@ -0,0 +1,811 @@
...
+
+        @Query("where price < :maxPrice order by title")
+        List<Book> cheaperThan(BigDecimal maxPrice);
+    }
+}
+----
+<1> This is the new part: `implements PanacheEntity.Stateless` gives our `Book` class methods to create, delete or update books without the need to use specific methods in repositories.
+
+<2> The repository also changed to `PanacheRepository.Stateless<Book, Long>`.
+// TODO: PanacheRepository<Book> infers Long, but Stateless/Managed/Reactive all require <Book, Long> explicitly. We should add single-parameter overloads (defaulting to Long) for consistency.
```
FroMage: This is fixed in the rename branch, they all default to `Long` unless `.CustomId`

_guides/quarkus-data-hibernate-getting-started.adoc
FroMage: I'm not sure. It explains to Hibernate users why this is different and tells them where to find what they're used to elsewhere.  Perhaps it's too long in details. Perhaps it should be titled "note for hibernate frequent users" (not sure about what term to use here).

_guides/quarkus-data-hibernate-getting-started.adoc
FroMage: There's quite a section about custom IDs already in the Quarkus Data guide (reference or not), so you could resume this entire section into the pointer.

_guides/quarkus-data-hibernate-getting-started.adoc
```diff
@@ -0,0 +1,811 @@
...
+[source,java]
+----
+package org.acme;
+
+import io.quarkus.hibernate.panache.PanacheRepository;
+import io.quarkus.hibernate.panache.WithId;
+import jakarta.persistence.Entity;
+
+@Entity
+public class Publisher extends WithId.AutoLong {
```
FroMage:
```suggestion
public class Publisher extends RecordEntity {
```

_guides/quarkus-data-hibernate-getting-started.adoc
```diff
@@ -0,0 +1,811 @@
...
+    public String name;
+    public String country;
+    public boolean allowsDiscounts;
+
+    public interface Repo extends PanacheRepository.Stateless<Publisher, Long> {
```
FroMage:
```suggestion
    public interface Repo extends RecordRepository<Publisher> {
```

_guides/quarkus-data-hibernate-getting-started.adoc
```diff
@@ -0,0 +1,811 @@
...
+    }
+}
+----
+
+And add a `publisher` relationship to `Book`, plus a stateless repository and a query to find books by publisher country:
+
+[source,java]
+----
+@Entity
+public class Book extends WithId.AutoLong implements PanacheEntity.Stateless {
```
FroMage:
```suggestion
public class Book extends StatelessEntity {
```

_guides/quarkus-data-hibernate-getting-started.adoc
```diff
@@ -0,0 +1,811 @@
...
+    public String title;
+    public String isbn;
+    public BigDecimal price;
+    @ManyToOne(fetch = FetchType.LAZY) // <1>
+    public Publisher publisher;
+
+    public interface Repo extends PanacheRepository.Stateless<Book, Long> {
```
FroMage:
```suggestion
    public interface Repo extends StatelessRepository<Book> {
```

_guides/quarkus-data-hibernate-getting-started.adoc
FroMage: Indeed, I thought fetches were explicit in stateless sessions?

_guides/quarkus-data-hibernate-getting-started.adoc
```diff
@@ -0,0 +1,811 @@
...
+        List<Book> modified = new ArrayList<>(); // <1>
+        BigDecimal factor = BigDecimal.ONE.subtract(discountPercent.divide(BigDecimal.valueOf(100)));
+        for (Book book : books) {
+            Publisher publisher = publisherRepo.findById(book.publisher.id); // <2>
+            if (publisher.allowsDiscounts) {
+                book.price = book.price.multiply(factor).setScale(2, RoundingMode.HALF_UP);
+                modified.add(book); // <3>
+            }
+        }
+        // Other business logic that modifies books i.e. applying loyalty points, rounding prices ...
```
FroMage, line 285: This is a contrived example, because in practice we're only updating each book once.

_guides/quarkus-data-hibernate-getting-started.adoc
```diff
@@ -0,0 +1,811 @@
...
+        }
+    }
+}
+----
+<1> We collect the books we modify so we can update them all at the end, after all business logic has run. Updating inside the body will result in multiple SQL update statements.
+<2> In a stateless session, lazy associations are not fetched automatically. We must load the publisher explicitly. A join fetch query in the Book repository could be used instead of a separate fetch, but for the scope of the tutorial we're not introducing join fetches.
+<3> We add the book to our list of modified objects.
+<4> We update once at the end, after all modifications are done.
+
+Notice how the persistence concerns, tracking which objects changed, fetching lazy associations, explicitly calling `update()`, are mixed in with the business logic. This method is beginning to become complex.
```
FroMage:
```suggestion
Notice how the persistence concerns, tracking which objects changed, fetching lazy associations, explicitly calling `update()`, are mixed in with the business logic. This method is becoming complex.
```

_guides/quarkus-data-hibernate-getting-started.adoc
```diff
@@ -0,0 +1,811 @@
...
+        // Prentice Hall (US, no discounts): prices unchanged
+        assertEquals(new BigDecimal("55.00"), bookRepo.findByTitle("Domain-Driven Design").getFirst().price);
+        assertEquals(new BigDecimal("50.00"), bookRepo.findByTitle("Refactoring").getFirst().price);
+    }
+}
+----
+<1> Rolls back the discounted prices after the test, so other tests and re-runs always see the original data from `import.sql`.
+
+This works, but it can be simpler.
+Quarkus Data Hibernate supports a "managed persistence context" that tracks objects for you: Hibernate monitors every entity it loads, detects which fields changed, and generates the minimal SQL needed to keep entities synchronized with the database at commit time. You don't need to track dirty objects, and lazy associations are fetched automatically when you access them.
```
FroMage, line 393: I feel like this is a very long introduction where we show a lot of code just to justify the existence of managed sessions.

I think we should rather make a list of pros/cons of each session type. That would be more informative and to the point, and remove a lot of code that is intended to be thrown away.

_guides/quarkus-data-hibernate-getting-started.adoc
```diff
@@ -0,0 +1,811 @@
...
+
+Switch `Book` to use `PanacheEntity.Managed` and a managed repository:
+
+[source,java]
+----
+@Entity
+public class Book extends WithId.AutoLong implements PanacheEntity.Managed { // <1>
```
FroMage:
```suggestion
public class Book extends ManagedEntity { // <1>
```

_guides/quarkus-data-hibernate-getting-started.adoc
```diff
@@ -0,0 +1,811 @@
...
+    public String title;
+    public String isbn;
+    public BigDecimal price;
+    @ManyToOne(fetch = FetchType.LAZY)
+    public Publisher publisher;
+
+    public interface Repo extends PanacheRepository<Book> { // <2>
```
FroMage:
```suggestion
    public interface Repo extends ManagedRepository<Book> { // <2>
```

_guides/quarkus-data-hibernate-getting-started.adoc
```diff
@@ -0,0 +1,811 @@
...
+            if (book.publisher.allowsDiscounts) { // <1>
+                book.price = book.price.multiply(factor).setScale(2, RoundingMode.HALF_UP); // <2>
+            }
+        }
+    } // <3>
+}
+----
+<1> Accessing `book.publisher` triggers a lazy fetch: Hibernate loads the publisher from the database automatically with a query.
+<2> We just set the field. No need to call `update()`.
+<3> When the `@Transactional` method returns, Hibernate compares every managed entity against its original state and generates the correct SQL. Books whose publisher didn't allow discounts were never changed, so Hibernate skips them entirely.
```
FroMage: I think this is not true, and scary: it does not compare every managed entity, I think it tracks those that were modified, no? If true, this is less scary because obviously more efficient.

_guides/quarkus-data-hibernate-getting-started.adoc
FroMage: I'm not sure that N+1 issues are related to "not immediately hitting the DB" and this may appear to suggest that 🤔

_guides/quarkus-data-hibernate-getting-started.adoc
```diff
@@ -0,0 +1,811 @@
...
+
+`quarkus-hibernate-reactive` adds the reactive Hibernate engine, which uses non-blocking I/O under the covers instead of JDBC.
+`quarkus-reactive-pg-client` is the reactive equivalent of `quarkus-jdbc-postgresql`. It provides a non-blocking PostgreSQL driver built on Vert.x. As its blocking counterpart, it's mandatory for Quarkus Data Hibernate to work.
+
+Now let's change only the minimal part of the code you need to try reactive. Let's add a second nested repository for reactive operations:
+
+[source,java]
+----
+@Entity
+public class Book extends WithId.AutoLong implements PanacheEntity.Managed {
```
FroMage:
```suggestion
public class Book extends ManagedEntity {
```

_guides/quarkus-data-hibernate-getting-started.adoc
```diff
@@ -0,0 +1,811 @@
...
+    public String title;
+    public String isbn;
+    public BigDecimal price;
+    @ManyToOne(fetch = FetchType.LAZY)
+    public Publisher publisher;
+
+    public interface Repo extends PanacheRepository<Book> { // <1>
```
FroMage:
```suggestion
    public interface Repo extends ManagedRepository<Book> { // <1>
```

_guides/quarkus-data-hibernate-getting-started.adoc
```diff
@@ -0,0 +1,811 @@
...
+        List<Book> findByTitle(String title);
+
+        @Query("where price < :maxPrice order by title")
+        List<Book> cheaperThan(BigDecimal maxPrice);
+
+        @Query("where publisher.country = :country")
+        List<Book> findByPublisherCountry(String country);
+    }
+
+    public interface ReactiveRepo extends PanacheRepository.Reactive<Book, Long> { // <2>
```
FroMage:
```suggestion
    public interface ReactiveRepo extends ManagedRepository.Reactive<Book> { // <2>
```

_guides/quarkus-data-hibernate-getting-started.adoc
FroMage: Absolutely, we should improve this.

FroMage (changes requested):

_guides/quarkus-data-hibernate-getting-started.adoc
yrodiere: It's not true in Quarkus because we use bytecode enhancement, which allows inline dirty-tracking.

It's true in vanilla Hibernate ORM: that's called dirty-checking.

FroMage: You know, I never realised these two terms were different :-/

_guides/quarkus-data-hibernate-getting-started.adoc
FroMage: Come to think about it, I wonder if we should not use the same sort of mechanism as I recently did to allow returning a `Uni` from `@Startup` and just plug it in the vertx loop. That would make for much nicer test methods.

lucamolteni:
So I've solved most of the issue but to acknowledge comment https://github.com/quarkusio/quarkusio.github.io/pull/2784#discussion_r3472729375 I will split this guide in two tutorials. I'll remove the SQL part from this one so that we can keep the discussion thread and merge this and I'll create a new one for the SQL part ok? 

@yrodiere @FroMage

_guides/quarkus-data-hibernate-getting-started.adoc
lucamolteni: Moved this part the SQL tutorial instead

_guides/quarkus-data-hibernate-getting-started.adoc
lucamolteni: Removed this part from this tutorial, moved it to the SQL tutorial

_guides/quarkus-data-hibernate-getting-started.adoc
lucamolteni: Moved to the SQL tutorial

_guides/quarkus-data-hibernate-getting-started.adoc
lucamolteni: Changed from ManyToOne to OneToMany so it's lazy by default

_guides/quarkus-data-hibernate-getting-started.adoc
lucamolteni: Replaced with ToMany association

_guides/quarkus-data-hibernate-getting-started.adoc
lucamolteni: Yes that's the whole point. Update each book only once, and only those that were actually changed

_guides/quarkus-data-hibernate-getting-started.adoc
lucamolteni: it fits the tutorial in which we just need to see if everything works and to see compiler errors when repositories are wrong

_guides/quarkus-data-hibernate-getting-started.adoc
lucamolteni: https://github.com/quarkusio/quarkus/issues/55651

_guides/quarkus-data-hibernate-getting-started.adoc
lucamolteni: Removed, though I personally find domain examples make tutorials more engaging than abstract ones

yrodiere:
Uh. @lucamolteni I just realized this is a PR to `quarkusio/quarkusio.github.io` instead of `quarkusio/quarkus`. Are you sure this is what you intended?

lucamolteni:
> Step-by-step guide covering Quarkus Data Hibernate repositories: entity mapping, repository pattern, managed vs stateless sessions, native SQL queries, and Hibernate Reactive.

I totally missed guides are supposed to be in the quarkus repositories. 

Replaced by https://github.com/quarkusio/quarkus/pull/55658

FroMage:
Nice way to start from a clean slate with zero comments you sneaky you 🤣

lucamolteni:
> Nice way to start from a clean slate with zero comments you sneaky you 🤣

I know it's meant to be a joke but I seriously included the original discussion to avoid the "blank slate" effect, in which an empty PR will attract a whole new set of comments :)

FroMage:
I know :) And I know it was not moved on purpose. I do find it a bit inconvenient that the comments are gone from the new PR, even if they are resolved (it helps reviewers check and not lose review history), but that's nothing we can control, that's the UX as it is 🤷

PR #2813, opened 2026-07-01
Add Lighthouse CI to PR build workflow to produce performance reports

holly-cummins:
This is something I think we ought to have, independent of migration.

As far as I can tell, Lighthouse, from Google Chrome, is still the industry standard for performance and best practices tests for sites. This PR runs Lighthouse against the test server after smoke tests, scoring performance, accessibility, best practices, and SEO for key pages. Results appear as a color-coded table in the GitHub step summary and are uploaded as build artifacts.

If we go below a certain threshold it will warn. For some categories like accessibility, it will fail the build, but otherwise, people need to check the reports to see the warnings. If the reports are useful we should maybe add them to the bot comments on the PR, but maybe we should wait to see if we like them first.

As with the e2e testing, it will run everything for infra changes, and otherwise it will just run changed pages. This does create a risk that a change a to a bad page would have bad scores and error because there aren't other pages to drag average scores up. We may need to wait and see how that plays out.

I was uneasy about supply chain issues, but lhci is owned by Chrome, and this uses npm ci with a lockfile to pin @lhci/cli dependencies. So it should be ok. 

Here's the report we get at the moment:
image

... and the full details live at https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1782923375367-10880.report.html. Looking through the report, there's clearly some perf work we should be doing. 

cc @insectengine

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,6 +10,13 @@ updates:
     schedule:
       interval: weekly
     open-pull-requests-limit: 20
+  - package-ecosystem: npm
+    directory: "/"
+    schedule:
+      interval: weekly
+    open-pull-requests-limit: 5
+    labels:
+      - area/infra
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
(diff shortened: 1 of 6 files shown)


insectengine (commented):
I can't speak the technical code but I agree with the concept. ;)

PR #2814, opened 2026-07-01
Updated git ignore

insectengine:
Updated .gitignore file to not include _context and agents.md files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,7 @@ CLAUDE.local.md
 # Blog preview marker files
 .blog-preview-last-run
 .blog-preview.lock
+
+# AI assistant context files
+AGENTS.md
+_context/


holly-cummins:
I don't know for sure what to do here, but there are some counter-arguments against these files:

https://www.augmentcode.com/blog/your-agents-context-is-a-junk-drawer

> In February 2026, researchers at ETH Zurich published [a paper evaluating AGENTS.md files](https://arxiv.org/abs/2602.11988) across multiple coding agents and LLMs. The finding was blunt.
>
>    Context files reduce task success rates compared to providing no repository context, while increasing inference cost by over 20%.
>
> Adding context files made agents perform worse than giving them nothing. And it cost more.
>
> The paper's author [clarified on Hacker News](https://news.ycombinator.com/item?id=43176795) that even human-written context files only improved performance by about 4%, and that improvement wasn't consistent across models. On Sonnet 4.5, performance actually dropped by over 2%.

Changing the git ignore isn't the same as adding the file, and so we possibly should change the gitignore even if we never add the file.

holly-cummins:
Please ignore my comments above, I read the PR backwards. 😳

PR #2815, opened 2026-07-02
Update perf graphics to use 'latest'

holly-cummins:
We haven't seen anything hideously unstable in the published benchmarks graphics, and updated the link on the front page is kind of tedious. We said after a period of stabilisation we'd switch to latest, so I propose that time is now.


holly-cummins:
Checked preview, 
image

Dark mode is also fine, unlike last time I tried to update the graphics!

edeandrea (approved):

PR #2827, opened 2026-07-09
Image Alt tab fix...

insectengine:
Based on the lighthouse report on website performance, I've used Bob to help find and correct missing alt tag elements.

1) Some blank alt tags are blank so they will be skipped by screen readers. This will prevent redundancy in the reader... " "Developer Joy icon, Developer Joy" as the heading immediately duplicates the alt.

2) Some of the changes affect includes that aren't currently in use. I haven't removed them as they maybe the basis for future content.


holly-cummins:
Oh nice, I meant to raise issues for some of the lighthouse findings but didn't get round to it!

holly-cummins (approved):

PR #2831, opened 2026-07-10
Fix duplicate tags and dead links on front page caused by missing downcase

holly-cummins:
Another one caught by https://github.com/quarkusio/quarkusio.github.io/pull/2713. @insectengine caught the problem independently, I think from looking at the lighthouse reports, so I think it affects our SEO or accessibility scores. 

On the homepage at the moment, because some blogs use tags like AI or LLM, there are dead links right on the front page:

image

Clicking on AI takes you to a 404. On the blog page, it's lower case:
image

### DIFF
--- a/_includes/recent-posts-band.html
+++ b/_includes/recent-posts-band.html
@@ -13,7 +13,7 @@
             <p class="datetag margin-tb-sm">
               {{ post.date | date: '%B %d, %Y' }} &nbsp;&nbsp;&nbsp;Tags:
               {% for tag in post.tags %}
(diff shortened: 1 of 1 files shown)


holly-cummins:
James and I are also hopeful this will fix duplicate tags.

insectengine (changes requested):
I'm still seeing multiples of the "ai" tag but the header is upper case with "AI"

holly-cummins:
> I'm still seeing multiples of the "ai" tag but the header is upper case with "AI"

I, uh, removed half the fix while fixing formatting. Oops.

Working now, I think!

insectengine (approved):

PR #2832, opened 2026-07-10
Change "insights" to "quarkus-insights" and remove tag of "quarkus"

insectengine:
For discoverability and clarity, I'm changing the "insights" tag to "quarkus-insights" on 17 posts. I'm also removing an orphan "quarkus" tag that doesn't need to exist.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,6 +10,13 @@ updates:
     schedule:
       interval: weekly
     open-pull-requests-limit: 20
+  - package-ecosystem: npm
+    directory: "/"
+    schedule:
+      interval: weekly
+    open-pull-requests-limit: 5
+    labels:
+      - area/infra
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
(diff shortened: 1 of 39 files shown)


holly-cummins:
I was briefly puzzled by the failures in this PR, then I worked out what was happening. It looks like the checks of #2713, but #2713 isn't merged. And I didn't know why #2713 was finding non-fragment failures. I assumed it had accidentally fixed a bug in the crawler. 

But no. It's that we happen to have blogs on the front page now (ie recent blogs) which expose the pre-existing tag problem fixed by #2831. This PR causes the link checker to run on the front page, which it wouldn't always, so it is failing with the issue.

holly-cummins (approved):
If we wanted to be really fancy we'd add a redirect from /tags/insights to /tags/quarkus-insights, but I don't know if anything would ever have linked to the old url.

insectengine:
Was easier to just close this PR and redo the work and resubmit to avoid the unnecessary commits.

PR #2849, opened 2026-07-14
Add Roq-compatible Java classes to replicate Jekyll plugins

holly-cummins:
Most of the quarkus.io site can be converted using scripts, but converting the Jekyll plugins to Roq is really hard to do. Claude's first attempts at scripts were basically a script that hardcoded in the new Java file and output it. :) 

The plugins also don't change much, so I think it's best to do the conversion as a one-off. I think we're close enough to cutting over now that we can risk the dual-maintenance of having both sets of code in the repo. The Java code is harmless, but also not being exercised at the moment. (I know it does work, because I have tests wrapping a lot of the function, and I've also visually verified in https://quarkus-website-pr-2683-preview.surge.sh/). 

Roq doesn't have a formal plugin mechanism, and instead, code just goes into the src folder. This code replicates 
- https://github.com/quarkusio/quarkusio.github.io/blob/main/_plugins/asciidoctor-extension.rb
- https://github.com/quarkusio/quarkusio.github.io/blob/main/_plugins/copy-search-wc.rb

The regex filter and strings plugins aren't needed, and the cname plugin is so standard I'm handling it in the converter. 

Asking @michalvavrik for a review since he's touched these classes most recently, although a review might need more knowledge of Asciidoctor-Java than any of us have. Also cc @marko-bekhta for awareness. 

Having the code in the repo in advance of the conversion doesn't _really_ change anything, but it makes it easier to review by not mixing it in with 42k changed files from the automated run. Although I guess the knowledge issue mentioned above still makes it hard to review.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,7 +30,7 @@ jobs:
           else
             echo "jekyll=false" >> $GITHUB_OUTPUT
           fi
-          if [ -d "src/main" ]; then
+          if [ -f "config/application.properties" ]; then
             echo "roq=true" >> $GITHUB_OUTPUT
             echo "site_dir=target/roq" >> $GITHUB_OUTPUT
           else
--- a/pom.xml
+++ b/pom.xml
@@ -10,24 +10,75 @@
         <maven.compiler.release>21</maven.compiler.release>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
+        <quarkus.platform.artifact-id>quarkus-bom</quarkus.platform.artifact-id>
+        <quarkus.platform.group-id>io.quarkus.platform</quarkus.platform.group-id>
+        <quarkus.platform.version>3.37.2</quarkus.platform.version>
+        <skipITs>true</skipITs>
         <surefire-plugin.version>3.5.6</surefire-plugin.version>
     </properties>
+
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>${quarkus.platform.group-id}</groupId>
+                <artifactId>${quarkus.platform.artifact-id}</artifactId>
+                <version>${quarkus.platform.version}</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
+
     <dependencies>
         <dependency>
-            <groupId>org.junit.jupiter</groupId>
-            <artifactId>junit-jupiter</artifactId>
-            <version>6.1.2</version>
-            <scope>test</scope>
+            <groupId>io.quarkiverse.roq</groupId>
+            <artifactId>quarkus-roq</artifactId>
+            <version>2.1.5</version>
         </dependency>
         <dependency>
-            <groupId>com.microsoft.playwright</groupId>
-            <artifactId>playwright</artifactId>
-            <version>1.61.0</version>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-arc</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkiverse.roq</groupId>
+            <artifactId>quarkus-roq-testing</artifactId>
+            <version>2.1.5</version>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkiverse.roq</groupId>
+            <artifactId>quarkus-roq-plugin-aliases</artifactId>
+            <version>2.1.5</version>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkiverse.roq</groupId>
+            <artifactId>quarkus-roq-plugin-tagging</artifactId>
+            <version>2.1.5</version>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkiverse.roq</groupId>
+            <artifactId>quarkus-roq-plugin-asciidoc-jruby</artifactId>
+            <version>2.1.5</version>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkiverse.playwright</groupId>
+            <artifactId>quarkus-playwright</artifactId>
+            <version>2.3.7</version>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-junit</artifactId>
             <scope>test</scope>
         </dependency>
     </dependencies>
+
     <build>
         <plugins>
+            <plugin>
+                <groupId>${quarkus.platform.group-id}</groupId>
+                <artifactId>quarkus-maven-plugin</artifactId>
+                <version>${quarkus.platform.version}</version>
+                <extensions>true</extensions>
+            </plugin>
             <plugin>
                 <artifactId>maven-compiler-plugin</artifactId>
                 <version>${compiler-plugin.version}</version>
@@ -40,10 +91,36 @@
                 <version>${surefire-plugin.version}</version>
                 <configuration>
                     <excludedGroups>e2e</excludedGroups>
+                    <argLine>@{argLine}</argLine>
+                    <systemPropertyVariables>
+                        <java.util.logging.manager>org.jboss.logmanager.LogManager</java.util.logging.manager>
+                        <maven.home>${maven.home}</maven.home>
+                    </systemPropertyVariables>
+                </configuration>
+            </plugin>
+            <plugin>
+                <artifactId>maven-failsafe-plugin</artifactId>
+                <version>${surefire-plugin.version}</version>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>integration-test</goal>
+                            <goal>verify</goal>
+                        </goals>
+                    </execution>
+                </executions>
+                <configuration>
+                    <argLine>@{argLine}</argLine>
+                    <systemPropertyVariables>
+                        <native.image.path>${project.build.directory}/${project.build.finalName}-runner</native.image.path>
+                        <java.util.logging.manager>org.jboss.logmanager.LogManager</java.util.logging.manager>
+                        <maven.home>${maven.home}</maven.home>
+                    </systemPropertyVariables>
                 </configuration>
             </plugin>
         </plugins>
     </build>
+
     <profiles>
         <profile>
             <id>e2e</id>
@@ -58,6 +135,19 @@
                     </plugin>
                 </plugins>
             </build>
+        </profile>
+        <profile>
+            <id>native</id>
+            <activation>
+                <property>
+                    <name>native</name>
+                </property>
+            </activation>
+            <properties>
+                <quarkus.package.jar.enabled>false</quarkus.package.jar.enabled>
+                <skipITs>false</skipITs>
+                <quarkus.native.enabled>true</quarkus.native.enabled>
+            </properties>
         </profile>
     </profiles>
 </project>
--- a/src/main/java/io/quarkus/tools/migration/asciidoc/ConfigTablePostprocessor.java
+++ b/src/main/java/io/quarkus/tools/migration/asciidoc/ConfigTablePostprocessor.java
@@ -1,0 +1,50 @@
+package io.quarkus.tools.migration.asciidoc;
+
+import org.asciidoctor.ast.Document;
+import org.asciidoctor.extension.Postprocessor;
+
+public class ConfigTablePostprocessor extends Postprocessor {
+
+    @Override
+    public String process(Document document, String output) {
+        Object maxIdObj = document.getAttribute("quarkus-config-max-row-id");
+        if (maxIdObj == null) {
+            return output;
+        }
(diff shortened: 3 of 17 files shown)


src/main/java/io/quarkus/tools/migration/SearchScriptDownloader.java
```diff
@@ -0,0 +1,41 @@
...
+import io.quarkus.runtime.Startup;
+import jakarta.enterprise.context.ApplicationScoped;
+import org.jboss.logging.Logger;
+
+@ApplicationScoped
+@Startup
+public class SearchScriptDownloader {
+
+    private static final Logger LOG = Logger.getLogger(SearchScriptDownloader.class);
+    private static final String SCRIPT_URL = "https://search.quarkus.io/static/bundle/app.js";
```
marko-bekhta: Maybe we should have an application.properties + config for this ?
We had a `direct` mode for local dev when you'd point the url to your local instance of search.quarkus.io and then easily modify the search components...

src/main/java/io/quarkus/tools/migration/SearchScriptDownloader.java
```diff
@@ -0,0 +1,41 @@
...
+    private static final String CACHED_FILE = "assets/javascript/search-wc.js";
+
+    // Downloads to public/ which survives mvn clean (only target/ is deleted).
+    // TODO: should mvn clean also remove this downloaded file?
```
marko-bekhta: I suppose if we can get that direct mode for the script this won't matter much: on CI the file wouldn't be there, and on local env one would use the direct mode -- so no download at all.

michalvavrik:
> Asking @michalvavrik for a review since he's touched these classes most recently, although a review might need more knowledge of Asciidoctor-Java than any of us have.

I am surprised that I am still the last one, I thought @gsmet and @ia3andy  did little tweaks here and there. Quite frankly, the last thing I remember was that I wrote it in ruby.

All that code seems familiar, I am happy to review, but to provide any real value, I'll need to find a time, so you would have to wait for a weekend or bit later.

I won't be offended  if you find quicker reviewers :-)

src/main/java/io/quarkus/tools/migration/asciidoc/TooltipInlineMacroProcessor.java
```diff
@@ -0,0 +1,29 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+import org.asciidoctor.ast.PhraseNode;
+import org.asciidoctor.ast.StructuralNode;
+import org.asciidoctor.extension.InlineMacroProcessor;
+import org.asciidoctor.extension.Name;
+
+@Name("tooltip")
+public class TooltipInlineMacroProcessor extends InlineMacroProcessor {
```
marko-bekhta, line 15:
```suggestion
@ContentModel(ContentModel.RAW)
public class TooltipInlineMacroProcessor extends InlineMacroProcessor {
```

I think this would be a closest alternative to the `resolve_attributes false` ?
and with it, the attribute would have a `text` attribute instead of the positional one.

If  we don't have it, then the `[ .. ]` would be parsed as attributes and any comma in there would create another positional attribute 🫣

src/main/java/io/quarkus/tools/migration/asciidoc/TooltipInlineMacroProcessor.java
```diff
@@ -0,0 +1,29 @@
...
+
+    @Override
+    public PhraseNode process(StructuralNode parent, String target, Map<String, Object> attributes) {
+        String text = attributes.get("1") != null ? attributes.get("1").toString() : "";
```
marko-bekhta:
```suggestion
        String text = Objects.toString(attributes.get("text"), "");
```
maybe ?

marko-bekhta (commented):
hey 🙂 
I looked through a few of processors and they look good overall, only thing I found missing so far was `resolve_attributes false`.

btw for the Hibernate extensions we have a small adoc in a test that helps debug  different processors, maybe we can add a test or some public-static-main here so it would be easy to run a quick test ?

src/main/java/io/quarkus/tools/migration/SearchScriptDownloader.java
holly-cummins: About that. 

It seems that kind of arbitrary config is the one thing that's really hard to do with Roq - I raised https://github.com/quarkiverse/quarkus-roq/issues/1083 to track. So I've been scratching my head over how to make the search endpoint be different in dev mode, which is something that's trivially easy in Jekyll. 

But you're right that here, this could totally be config, because it's Java! And while I'm there I *probably* can even be able to extend it to handle the config for the search endpoint, with the only question being how hacky it is. So it's a good spot about the hard-coding, thanks, and I think you've shown me how to fix another problem, extra-thanks!

src/main/java/io/quarkus/tools/migration/search/SearchConfig.java
```diff
@@ -0,0 +1,46 @@
+package io.quarkus.tools.migration.search;
+
+import io.smallrye.config.ConfigMapping;
+import io.smallrye.config.WithDefault;
+
+@ConfigMapping(prefix = "search")
+public interface SearchConfig {
```
marko-bekhta: 🥳 🙂

holly-cummins: I'm testing this locally (which is complex) and what I have isn't quite right, so I'll do another update.

holly-cummins:
Ok, I think this is ready for review. Well, as ready as code which is disconnected from action can be. I've got the config in, and I'm running with these files in my local roq conversion, and things seem to be working. Thank you for the excellent suggestions and fixes, @marko-bekhta!

pom.xml
```diff
@@ -1,33 +1,91 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
-    <groupId>io.quarkusio</groupId>
-    <artifactId>quarkusio-website-tests</artifactId>
+    <groupId>io.acme</groupId>
+    <artifactId>roq-create-62704</artifactId>
```
michalvavrik: This is hard to understand. I don't have a context, but if there is a good reason to turn `io.quarkusio:quarkusio-website-tests` into `io.acme:roq-create-62704`, it requires a good name as the latter is incomprehensible to me.

michalvavrik (commented):
I think PR looks great and other reviewers have it covered. I am terribly busy and probably can't be useful anyway. I'm going quiet.

pom.xml
holly-cummins: You're right, @michalvavrik, there's no good reason for this. It came across from an automated roq `create`, but it shouldn't be there. Will fix, thank you!

holly-cummins: In fact, the only pom changes that should be there are the addition of a few dependencies.

holly-cummins:
Oops, I did a conversion on this branch by accident. Fixed now!

.github/workflows/build.yml
```diff
@@ -30,7 +30,7 @@ jobs:
           else
             echo "jekyll=false" >> $GITHUB_OUTPUT
           fi
-          if [ -d "src/main" ]; then
+          if [ -f "config/application.properties" ]; then
```
marko-bekhta, line 33: just to double-check -- this config is created during the "migration" right?

src/main/java/io/quarkusio/search/config/SearchConfig.java
```diff
@@ -0,0 +1,56 @@
...
+    int moreTimeout();
+
+    /**
+     * The minimum number of characters before we run a full search.
+     * Below this:
+     * - if another filter is selected (e.g. categories), we run Javascript search
+     * - otherwise, we don't run search and just display all guides
+     */
+    @WithDefault("2")
+    int minChars();
```
marko-bekhta, line 55: I think these 3 are only for the `<qs-form>` attributes, or do we need them for the config to load correctly?

src/main/java/io/quarkus/tools/migration/asciidoc/ConfigTablePostprocessor.java
```diff
@@ -0,0 +1,55 @@
...
+    public String process(Document document, String output) {
+        String docId = document.getId();
+        if (docId == null || !docId.startsWith("page-with-config-")) {
+            return output;
+        }
+
+        Object maxIdObj = document.getAttribute("quarkus-config-max-row-id");
+        if (maxIdObj == null) {
+            return output;
+        }
```
marko-bekhta:
```suggestion
        Object maxIdObj = document.getAttribute("quarkus-config-max-row-id");
        if (maxIdObj == null) {
            return output;
        }
```
looks like we set both id and this attribute in `ConfigTableTreeprocessor` so maybe we can skip the first check since we do not use the id further down and just check the max-row-id value ?

src/main/java/io/quarkus/tools/migration/asciidoc/ConfigTablePostprocessor.java
```diff
@@ -0,0 +1,55 @@
...
+            String cssClass = "";
+            if (id > 30) {
+                cssClass = " row-hidden";
+            }
+            if (id % 2 == 1) {
+                cssClass += " odd";
+            }
+
+            String collapsibleMarker = "id=\"conf-collapsible-desc-" + id + "\"";
+            int descIdx = sb.indexOf(collapsibleMarker);
```
marko-bekhta, line 28: btw maybe if we have some `int previouslyFoundIndex=0;` before the loop and we update it as we go, we could speed up the lookups of these collapsibleMarker/nonCollapsibleMarker a bit, as we wouldn't need to start from the beginning?

But that's probably for a later optimisation (I'm not sure how fast these already are on all the docs) 🙂

src/main/java/io/quarkus/tools/migration/asciidoc/ConfigTableTreeprocessor.java
```diff
@@ -0,0 +1,108 @@
...
+                    Block newCaption = createBlock(parent, "paragraph",
+                            captionContent + " <input type=\"search\" id=\"config-search-%d\" placeholder=\"FILTER CONFIGURATION\" disabled>".formatted(searchFieldId),
+                            new HashMap<>());
+                    siblings.set(tableIndex - 1, newCaption);
+                }
+            }
+        }
+
+        if (rowId > 0) {
+            document.setId("page-with-config-0");
```
marko-bekhta: maybe ?
```suggestion
```
that is if it's ok to remove the `!docId.startsWith("page-with-config-")` check in the post processor 🤞🏻 
I'll admit that this id 0 confused me 🙂

src/main/java/io/quarkus/tools/migration/search/SearchScriptDownloader.java
```diff
@@ -0,0 +1,49 @@
...
+
+    private static final Logger LOG = Logger.getLogger(SearchScriptDownloader.class);
+
+    @Inject @Named("searchConfig")
+    SearchConfig searchConfig;
+
+    // Downloads to public/ which survives mvn clean (only target/ is deleted).
+    @PostConstruct
+    void downloadScript() {
+        String scriptUrl = searchConfig.host() + searchConfig.scriptPath();
```
marko-bekhta: btw maybe we should return early and skip the download if the script mode is direct ? (but maybe log a message in that case..)

marko-bekhta (approved):
hey 👋🏻 🙂  I went trough the extensions comparing them to the ruby ones -- all looks nice! 🥳 

And thanks for adding those tests! I already debbuged a few of these extensiosn with them 🙂 !

I had some questions added inline, but those shouldn't block the pr 🫣

.github/workflows/build.yml
holly-cummins: It is, yes. Reviewing is always hard and annoying, and this one is extra hard and annoying because you're reviewing something which is built against a 'ghost' set of code that isn't in this repo yet. :(

pom.xml
holly-cummins: Hmm, I thought I'd fixed this, but the affected file section is in the PR.

src/main/java/io/quarkusio/search/config/SearchConfig.java
holly-cummins: Anything that varies by profile needs to go into the config mapping. I'm not sure these three are likely to change with the profile. Well, it seems pretty unlikely they would. But since they're so closely related to the others, I think it makes sense to declare them all in the same place.

src/main/java/io/quarkus/tools/migration/search/SearchScriptDownloader.java
holly-cummins: Makes total sense, done

PR #2861, opened 2026-07-21
Strengthen tags test and fix tagging regression

holly-cummins:
Oops. 
image

http://quarkus.io/blog has regressed, caused by my fix to the duplicate tags. What's extra annoying is it's working fine in Roq, so this is a very temporary fix, and getting Roq to tolerate my broken fix syntax was quite hard work.  Liquid applies filters left-to-right: push returns the array, then downcase is applied to the array (not the element), which stringifies and collapses it. The fix is to downcase the tag name first in a separate variable, then push it.

I wouldn't bother fixing since I hope to cut over soon, except that (a) it's doesn't look great and (b) I want to ship the strengthened test.

### DIFF
--- a/_includes/blog-band.html
+++ b/_includes/blog-band.html
@@ -68,7 +68,8 @@
     <h3 class="tags-label">Tags</h3>
     {% assign tag_keys = "" | split: "" %}
     {% for stats in site.tags %}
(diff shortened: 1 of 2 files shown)


holly-cummins:
Much better:
image

insectengine (approved):

insectengine:
Looks good to me!

PR #2862, opened 2026-07-21
Update contributor- and LLM-facing docs for Roq

holly-cummins:
Should be merged after #2683.

I did an automated migration of blog-preview.sh but did not test it.


holly-cummins:
Marking as draft so we don't accidentally merge this before #2683.

holly-cummins:
This should now go in. Until it does, I think LLMs will find our site confusing and be a bit stupider.

MichalMaler (approved):

PR #2865, opened 2026-07-23
Be more specific about which versions herodevs supports

holly-cummins:
https://www.herodevs.com/support/nes-for-quarkus shows that HeroDevs support 2.16 and 3.20, not all versions. I've updated the /releases table (and backing data) to reflect this. I've also updated the link, which was a guess (and a wrong guess), to the new one. 

https://quarkus-website-pr-2865-preview.surge.sh/releases/ shows the new page. the change is most visible when 'hide all releases' is unticked (I did a ridiculous zoom here to show it): 
image


Sanne:
Great, thanks @holly-cummins !

PR #2868, opened 2026-07-23
Add blog post: To Cache or Not to Cache Virtual Threads

franz1981:
Investigation into pooling virtual threads


_posts/2026-07-23-to-cache-or-not-to-cache-virtual-threads.adoc
```diff
@@ -0,0 +1,262 @@
+---
+layout: post
+title: "To Cache or Not to Cache Virtual Threads"
+date: 2026-07-23
```
Sanne: 28
```suggestion
date: 2026-07-28
```

_posts/2026-07-23-to-cache-or-not-to-cache-virtual-threads.adoc
```diff
@@ -0,0 +1,262 @@
...
+Creating a virtual thread is cheap — have millions, and don't pool them!
+____
+
+image::meme-bell-curve.png[Bell curve meme]
+
+As part of our work on https://youtu.be/Oy005l5vHtE[improving Quarkus performance with virtual threads] and https://youtu.be/oWE_kYB4Wns[investigating Loom's scheduling behavior], we decided to test this advice with data. *Spoiler: we should have listened.* But the _reason_ we should have listened turned out to be far more interesting than "they're cheap enough."
+
+== Why Would You Pool Virtual Threads?
+
+Pooling threads is a well-established pattern. We do it for platform threads in frameworks like Quarkus because they are expensive to create — each one maps to an OS thread with a pre-allocated stack. *Virtual threads are designed to be cheap, so the JDK authors say pooling is unnecessary.* But a brave and careless developer might still consider it: creating a virtual thread allocates a `Thread` object, its thread-local storage (a `HashMap` per thread), and internal scheduling structures. *If you reuse virtual threads, you skip all of that.* At a superficial analysis, this looks beneficial — the same reasoning we apply when pooling platform threads.
```
Sanne:
```suggestion
Pooling threads is a well-established pattern. We do it for platform threads in frameworks like Quarkus because they are expensive to create — each one maps to an OS thread with a pre-allocated stack. *Virtual threads are designed to be cheap, so the architects of Project Loom say pooling is unnecessary.* But a brave and careless developer might still consider it: creating a virtual thread allocates a `Thread` object, its thread-local storage (a `HashMap` per thread), and internal scheduling structures. If you reuse virtual threads, you skip all of that. At a superficial analysis, this looks beneficial — the same reasoning we apply when pooling platform threads.
```

_posts/2026-07-23-to-cache-or-not-to-cache-virtual-threads.adoc
```diff
@@ -0,0 +1,262 @@
...
+
+=== What could go wrong?
+
+*The workload we care about* is the typical Java enterprise pattern: an HTTP server receives requests, *dispatches each one to a virtual thread that performs blocking I/O* (a database query, an HTTP call to another service), then returns a response. This is how frameworks like Quarkus and Vert.x work — a Netty event loop accepts connections and hands off blocking work to virtual threads.
```
Sanne: Regarding "a Netty event loop accepts connections and hands off blocking work" .. I wonder if we should remind people of the role of Netty in Quarkus?

This section is quite brilliand as it's keeping things very simple though. An alternative would be to keep it simple but then don't confuse people with a "Netty event loop".

_posts/2026-07-23-to-cache-or-not-to-cache-virtual-threads.adoc
```diff
@@ -0,0 +1,262 @@
...
+
+image::quarkus-architecture.png[Quarkus architecture]
+
+*We've prepared a benchmark* which simplifies this to its essential shape: a Netty HTTP server receives requests, hands off each one to a VT that makes a blocking HTTP call (via Apache HttpClient) to a mock backend with 30ms think time, then returns the response. 10,000 concurrent connections, 8 cores, with a *30-second warmup phase* followed by a *30-second measurement phase*. In pooling mode, we pre-allocate a pool of 10,000 virtual threads — one per connection.
```
Sanne, line 37: Should we mention any limits by the Apache Httpclient? I assume you'd not allow it to open an unbounded number of concurrent connections?

_posts/2026-07-23-to-cache-or-not-to-cache-virtual-threads.adoc
```diff
@@ -0,0 +1,262 @@
...
+* https://github.com/openjdk/jdk/blob/master/src/hotspot/share/oops/instanceStackChunkKlass.hpp[Layout diagram in `InstanceStackChunkKlass.hpp`] (the chunk's internal structure)
+* https://github.com/openjdk/jdk/blob/master/src/hotspot/share/runtime/continuationFreezeThaw.cpp[Stack diagram in `continuationFreezeThaw.cpp`] (what gets frozen from the carrier stack)
+* https://github.com/openjdk/jdk/commit/9583e3657e43cc1c6f2101a64534564db2a9bd84[JDK-8284161] — the commit that introduced the freeze/thaw machinery
+====
+
+Two properties of StackChunks matter for our story:
+
+. *A chunk is sized exactly to fit the stack at freeze time.* Whatever stack depth the virtual thread has when it first parks determines the chunk's capacity — and that capacity never changes. *There is no resizing.*
+
+. *On each park, the JVM reuses the existing chunk* — it copies the stack into it via a fast bulk memory copy, no GC bookkeeping overhead. As long as the chunk stays in the *young generation* of the heap, this works perfectly. No new allocation needed.
```
Sanne: I'm a bit confused here - what if it gets parked a second time, and this second time it has a deeper chunk? I assume what you mean is that it's already resized the StackChunk and so will just use the same object instance again.
But this contradicts the previous statement " There is no resizing. "

I guess what you mean is, it's not truncated / made smaller at freeze time, in case the existing data structure is larger than strictly necessary, so it only grows?

_posts/2026-07-23-to-cache-or-not-to-cache-virtual-threads.adoc
```diff
@@ -0,0 +1,262 @@
...
+
+This design *prioritizes freeze/thaw speed over memory efficiency* — and for short-lived virtual threads, it works well. But for long-lived VTs, *every GC cycle that promotes a chunk triggers a fresh allocation and an abandonment cycle*. With 10,000 long-lived pooled VTs, every GC cycle:
+
+* Promotes 10,000 StackChunks to old gen
+* Triggers 10,000 fresh chunk allocations (new young-gen chunks)
+* Abandons 10,000 old-gen chunks — dead weight until the next Full GC cleans them up
+
+On a 1 GB heap, this churn was *severe enough to dominate GC behavior*.
+
+== If Pooling Doesn't Help, Pool More.
```
Sanne:
```suggestion
== When Pooling Doesn't Help, Pool More.
```

_posts/2026-07-23-to-cache-or-not-to-cache-virtual-threads.adoc
```diff
@@ -0,0 +1,262 @@
...
+== So... Don't Pool?
+
+It depends. With small heaps, clearly not. In production, you don't control VT lifecycles — and with a shared pool, you can't predict which code path a VT will be used for, whether that code is warm, hot, or never compiled by C2. Too aggressive caching risks bloating the runtime heap forever.
+
+But in Quarkus, we're experimenting with different options to improve our Loom integration — boldly going where no one has gone before. That means running with scissors that everyone told us are too sharp. We do it for the sake of innovation, and because we now know _why_ they cut. As the saying goes, *the journey matters more than the destination.*
+
+== Takeaways
+
+. *Pooling VTs doesn't save what you think it saves.* Under the hood, the JVM can still allocate new StackChunks — the biggest cost you were trying to avoid.
+. *The "don't pool" advice is correct.* The OpenJDK team was right.
```
Sanne: Well - you actually raised several further quesitons about this which go unanswered?

Firstly - I might have 4G of RAM and want to benefit from the higher throughput in your first graph: imagine I need those 143,966 TPS and I'm "fine" to ignore full-GC spikes (sure it's far fetched - but it's potentially valid in some contexts).

Second - Perhaps with some further work one would be able to benefit from this w/o the drawbacks?

Some ideas for potentially more experiments, (and more trouble):

1. have the ability to reduce the chunks (or perhaps better rotate them with a new copy strategically in certain situations, so to also benefit from young-gen positioning).
2. start the pooling only after steady state is reached (this would go well in hand with other optimisations you had in mind...)

I'm **not saying you should explore these first**. But this particular sentence is not such a strong takeway I'd say, as you've actually raised some doubts about it being an absolute truth.

cc/ @dmlloyd

Sanne (changes requested):
It's great and I think you could go live with it soon - have some minor remarks.

_posts/2026-07-23-to-cache-or-not-to-cache-virtual-threads.adoc
franz1981: > I guess what you mean is, it's not truncated / made smaller at freeze time, in case the existing data structure is larger than strictly necessary, so it only grows?

AFAIK if the stack chunk is capable to contain the stack frames and the chunk is yet "young", jdk mainline already allow to reuse it.
Clearly if young but cannot contains the new data, a new one need to replace the previous one.
In this regard a stack chunk is "immutable" in term of its available capacity, and won't grow.
That's why I said "there's no resizing" -but I see that need some few more words to make it clear :pray:

_posts/2026-07-23-to-cache-or-not-to-cache-virtual-threads.adoc
franz1981: On the https://github.com/openjdk/loom/pull/228 branch I have already suggested several directions i.e. a smarter reuse policy - as I believe is a solvable problem (native allocators do exactly this, with some degree of success .-. ), but I'm waiting to get some feedback, because my patch cover just for parallel but G1 is now (JDK 27) the default even for tiny/memory constrained apps.

_posts/2026-07-23-to-cache-or-not-to-cache-virtual-threads.adoc
Sanne: ok but this needs clarifying as I have the impression the two bullet points are in contradiction.

e.g. Now you said "clearly .. a new one needs .." but also "No new allocation needed."

_posts/2026-07-23-to-cache-or-not-to-cache-virtual-threads.adoc
Sanne: great. Yes I don't mean to suggest we should hint at such explorations - but I'd soften the language here as it's not an absolute truth as a takeaway.

```suggestion
. *The "don't pool" advice is broadly sensible.*
```

_posts/2026-07-23-to-cache-or-not-to-cache-virtual-threads.adoc
franz1981: We don't hit any of these limits so i didn't want to people to focus on that as not relevant :pray:

franz1981:
PTAL again @Sanne addressed the most of them (if not all)

_posts/2026-07-23-to-cache-or-not-to-cache-virtual-threads.adoc
```diff
@@ -0,0 +1,262 @@
...
+* https://github.com/openjdk/jdk/blob/master/src/hotspot/share/oops/instanceStackChunkKlass.hpp[Layout diagram in `InstanceStackChunkKlass.hpp`] (the chunk's internal structure)
+* https://github.com/openjdk/jdk/blob/master/src/hotspot/share/runtime/continuationFreezeThaw.cpp[Stack diagram in `continuationFreezeThaw.cpp`] (what gets frozen from the carrier stack)
+* https://github.com/openjdk/jdk/commit/9583e3657e43cc1c6f2101a64534564db2a9bd84[JDK-8284161] — the commit that introduced the freeze/thaw machinery
+====
+
+Two properties of StackChunks matter for our story:
+
+. *A chunk's capacity is fixed at allocation time.* It never grows or shrinks. If the frozen stack fits, the chunk is reused as-is — fast bulk memory copy, no GC overhead, no new allocation. If the stack is deeper than the chunk can hold, the JVM allocates a new chunk to replace it — but this is replacement, not resizing.
+
+. *This fast reuse only works while the chunk is in the young generation of the heap.*
```
Sanne, line 113: oh - why?

_posts/2026-07-23-to-cache-or-not-to-cache-virtual-threads.adoc
```diff
@@ -0,0 +1,262 @@
...
+With C1-only, the chunk capacity matches what we see in the pooled run under normal compilation. *The ~1,149-word chunks are C1-sized.* Under normal compilation, C2 shrinks the frozen stack to ~280 words — but the shared VT pool is created during warmup, before C2 kicks in. Its VTs are born with C1-sized chunks, and *that capacity is locked in forever.*
+
+The reuse fix keeps those bloated chunks alive in old gen — forever — leaving the garbage collector to fight over the scraps of a heap it can never fully reclaim.
+
+== So... Don't Pool?
+
+It depends. With small heaps, clearly not. In production, you don't control VT lifecycles — and with a shared pool, you can't predict which code path a VT will be used for, whether that code is warm, hot, or never compiled by C2. Too aggressive caching risks bloating the runtime heap forever.
+
+But in Quarkus, we're experimenting with different options to improve our Loom integration — boldly going where no one has gone before. That means running with scissors that everyone told us are too sharp. We do it for the sake of innovation, and because we now know _why_ they cut. As the saying goes, *the journey matters more than the destination.*
```
Sanne, line 244: 💯

Sanne (approved):

_posts/2026-07-23-to-cache-or-not-to-cache-virtual-threads.adoc
franz1981: That's the current state of the JDK work: the reason I can see is that it keeps the memcpy fast path lean and possible, including updating the chunk oops map (which are not needed for young gen chunks pointing to object references).
I think I wrote it somewhere in this article about it, at 

> When a long-lived virtual thread’s StackChunk gets promoted to old gen, the fast copy no longer works — writing into old-gen objects requires extra GC bookkeeping called [write barriers](https://shipilev.net/jvm/anatomy-quarks/13-intergenerational-barriers/) to track cross-generational references.

> So the JVM does the pragmatic thing: it allocates a new young-gen chunk and sets it as the new tail

That's the whole point of my PR: you have a slower path to copy from/to the stack and updating oops maps, BUT you save the new allocation to happen in exchange

PR #2894, opened 2026-08-05
Blog post feature image for social posting

insectengine:
@brunobat pointed out that when folks post links to our blog posts, they display our default card and are rather bland. Looking at adding the feature image in so that social posts will have content specific previews. 

This has a conditional on the base layout to post the default card when a feature image isn't present.


insectengine:
This really impacts the work done for the releases and I'd like to minimize extra work for others. 

The feature images can be created in a couple ways...
1) With coordination, I can create any images necessary.
2) Folks can self serve for the releases. I've supplied claude with a webp background image and an example image and a prompt that defined font and size and and wording and it produced one nicely.

holly-cummins:
+1, this will be great to have. Is this ready for review, @insectengine? It's marked as a draft.

Related to https://github.com/quarkusio/quarkusio.github.io/issues/1723, although that's for guides.

gsmet:
Hi,

I see where this is coming from and I can see the value of it but I really don't want to add more coordination burden to the releases. So either we generate images upfront for all the upcoming versions or we should have a look at what @gastaldi prepared here: https://github.com/quarkusio/quarkusio.github.io/pull/1725 .

And port if to Roq.

holly-cummins:
It would be nice to combine the two, so auto-generate if nothing is specified, but use what's given if a person defines one. That would give more variety than just using auto-generated images, that would probably all have the same background and format. 

What James has implemented now is "keep current default if nothing is specified, and use what's given if a person defines one." I think that's still an improvement over the current situation, as long as we thought the implementation would be compatible with #1725. Since #1725 will need a complete rewrite for Roq (plugins do NOT translate at all well), we could probably ensure it would be.

insectengine:
@gsmet - Added images for future releases through 3.40.5 to give us time to automate the process. If not I'll create them in batches to keep a head of the releases.

holly-cummins:
Oh dear, this missed the big Roq train, despite being a nice improvement, IMO. I'll run `roq it` on it to try to rescue it.

holly-cummins:
I think I've converted this to Roq (sorry, I got distracted), but while I was doing that, I think I noticed something non-ideal. The generated images (and most images we're likely to generate in future) don't add a lot to the post. We'd end up repeating the title three or four times at the top of the blog posts, once in an image and three times in text. 

So we maybe want to make a distinction between social images, which we'd always want, and top-of-post images, which we'd only want if they were something special, like a picture of Guillaume riding a unicycle while juggling bananas and machetes. Gatsby uses 'cover' for the thumbnail image (which would be a social one), and doesn't put it in the post. I don't know what conventions other systems use.

insectengine:
> I think I've converted this to Roq (sorry, I got distracted), but while I was doing that, I think I noticed something non-ideal. The generated images (and most images we're likely to generate in future) don't add a lot to the post. We'd end up repeating the title three or four times at the top of the blog posts, once in an image and three times in text.
> 
> So we maybe want to make a distinction between social images, which we'd always want, and top-of-post images, which we'd only want if they were something special, like a picture of Guillaume riding a unicycle while juggling bananas and machetes. Gatsby uses 'cover' for the thumbnail image (which would be a social one), and doesn't put it in the post. I don't know what conventions other systems use.

I'm good with using a different name here to separate the two uses. "Feature" for images at the top of posts and "Social" or "cover" for social specific images that are generated.

insectengine:
@holly-cummins - I changed "featureimage" to "socialimage" in the metadatat and I removed it from the post layout. I also created post images for releases through 3.43.5 (from .0 to .5).

PR #2900, opened 2026-08-07
WIP: Guide navigation / Guides by category / alternative search

marko-bekhta:
This is built on top of what @gsmet already had in 

- https://github.com/quarkusio/quarkusio.github.io/pull/2749

Opening it mostly so that people could click around and see where things are so far 🫣 🙂 

Since the all guides view is slightly tied to the search, it actually has to continue relying on those components built in the search service and even use new ones for the category navigation (on the all guides page). 

What else is in here:

- updated guide cards and layout of the Guides page (only `main` version)
- new way of searching
- updated guide page 
  - toc is now on the left
  - there's a "drawer" button on the veeeeeeery left that shows the list of all guides.
- there's a new config for dev -- only main/latest guides and no `--force_polling` since that one just kills CPU
- (ah and yeah -- switch to a staging env where corresponding search updates are currently deployed)
- (dark mode is not adjusted, no new var alternativbes there yet)
- probably something else I forgot about 🙂 

things I am not super happy about:
- building a json "on the page" to pass to the components (need the info about categories to build the categories view + search results)

inital:
image
search results:
image
Guide:
image
image


marko-bekhta:
I've also tried adding the pinned guides:
image

but not sure yet whether I like it 🙂 (now that it is in place)

marko-bekhta:
a couple updates: 
image
image
image

A version select has a "chevron" on a slightly toned-down background to make sure users can see that it is clickable.
Background on cards is gone, and I've added the same color for card titles on podcasts (maybe that can work ? give some color to pages so to speak 🙂) they both use an scss color variable so should match.
Some padding here and there (after the guide groups, after the titles in the "drawer" on a single guide page).

Edit: most likely requires a "hard reload" for things to get picked up.

insectengine:
@marko-bekhta @gsmet - Revised placement/sizing of version/search looks good across resolution sizes.
Screenshot 2026-08-17 at 2 13 01 PM

Screenshot 2026-08-17 at 2 11 12 PM
Screenshot 2026-08-17 at 2 10 57 PM

gsmet (approved):
LOVE IT \o/

insectengine (approved):
I say this is good to go!

marko-bekhta:
let me check how the new components/search work on the quarkiverse side (yeah, we use the same service/components there too). I'll attach some screenshots here when I have them 🤞🏻

marko-bekhta:
Here's how the new components look on the quarkiverse side:
image
vs existing:
image

holly-cummins:
Padding seems squished under the breadcrumb on all-config (which I was kind of surprised to see changed)

Before:

image

After:

image

holly-cummins:
Also, is it intentional that Main is in bold on the individual guide pages? 
image

holly-cummins:
Oh hang on, I'm looking at stale content pre-conversion, so I'll wait.

holly-cummins:
This isn't a blocker, because this looks SO GOOD and is such a huge improvement, but I find the word clouds to the right look a bit messy. I wonder if they'd look better with a super-pale border, just to give some structure? Or a super-pale background, but I think just the border. But even if @insectengine agrees with me, we could iterate on that.

insectengine:
> Also, is it intentional that Main is in bold on the individual guide pages? image

there isn't a reason for Main to be bolded.

insectengine:
> This isn't a blocker, because this looks SO GOOD and is such a huge improvement, but I find the word clouds to the right look a bit messy. I wonder if they'd look better with a super-pale border, just to give some structure? Or a super-pale background, but I think just the border. But even if @insectengine agrees with me, we could iterate on that.

Not sure where you're seeing the word clouds.

holly-cummins:
> > This isn't a blocker, because this looks SO GOOD and is such a huge improvement, but I find the word clouds to the right look a bit messy. I wonder if they'd look better with a super-pale border, just to give some structure? Or a super-pale background, but I think just the border. But even if @insectengine agrees with me, we could iterate on that.
> 
> Not sure where you're seeing the word clouds.

I was trying to come up with a name for this assemblage of words. I guess it's not technically a word cloud since the sizes are all the same, but the layout sort of reminds me of one:

image

insectengine:
As I look back into this, I see a consistency issue with the pinned "All configuration" page. While the page is full width with no TOC, it should still use the same title configuration as the rest of the pages. I currently has the titlebar in use as well. I'm attaching a mockup of how it should be configured to match the rest of the guide pages.

This page: https://quarkus-website-pr-2900-preview.surge.sh/version/main/guides/all-config

**Current page:** 
Screenshot 2026-08-25 at 3 02 15 PM

**Revised Mockup:** 
allconfig_refactor

insectengine:
> > > This isn't a blocker, because this looks SO GOOD and is such a huge improvement, but I find the word clouds to the right look a bit messy. I wonder if they'd look better with a super-pale border, just to give some structure? Or a super-pale background, but I think just the border. But even if @insectengine agrees with me, we could iterate on that.
> > 
> > 
> > Not sure where you're seeing the word clouds.
> 
> I was trying to come up with a name for this assemblage of words. I guess it's not technically a word cloud since the sizes are all the same, but the layout sort of reminds me of one:
> 
> image

Ah, ok. You mentioned it being on the right so I was doubly confused. The Category navigation, it's like a TOC. The two sizes of text are for Categories and Sub Categories. It visually lives in space of filters like on the Extensions page. Like on that page, we shouldn't have a surrounding visual element because the page is already full of cards.  This would add more visual clutter. 

That being said, this current version is a compromise to keep the list shorter so more categories are above the fold. Originally, I proposed it as flattened to one per line. Functionally, this jumps to anchors, whereas a true filter would reduce the number of cards.

gsmet:
I made the All configuration options page consistent here: https://github.com/quarkusio/quarkusio.github.io/pull/2938 .

holly-cummins:
> Ah, ok. You mentioned it being on the right so I was doubly confused.

Sorry! Do you know, I failed a driving test once for mixing up left and right? It turns out driving examiners are very picky about that kind of thing. 

> The Category navigation, it's like a TOC. The two sizes of text are for Categories and Sub Categories. It visually lives in space of filters like on the Extensions page. Like on that page, we shouldn't have a surrounding visual element because the page is already full of cards. This would add more visual clutter.

Yeah, I agree that more visual elements isn't ideal. I was wondering if something really really pale would hint at the underlying structure ('these are kind of tags', which becomes clear when they're selected), without adding too much clutter. 

I wonder, if it's a toc, if the selection behaviour should change? I wouldn't expect selecting an element in a TOC to make a tag-like lozenge. I don't know what I _would_ expect, maybe the other elements becoming lighter?

Even though I know intellectually that it's a list, because of the spacing and the fact that there's not much vertical alignment between the words (since they're all different lengths), it looks kind of like a pile of words to me. And then as soon as one is selected, it looks like tags for the selected element.

But not a blocker, as I said.

marko-bekhta:
going to close this one as we've merged the Roq version of it 🥳 :

https://github.com/quarkusio/quarkusio.github.io/pull/2950

PR #2919, opened 2026-08-24
Add adoc attributes override so dev mode guide xref links (and link tests) work

holly-cummins:
In dev mode, some xref guides links don't work. This also makes the link validation test fail in dev mode. See https://github.com/quarkiverse/quarkus-roq/pull/1158 for more discussion.

I think the nicest fix for this is in Roq, but we can work around it locally.

### DIFF
--- a/content/guides/_attributes-local.adoc
+++ b/content/guides/_attributes-local.adoc
@@ -5,4 +5,7 @@
 :code-examples: {generated-dir}/examples
 :imagesdir: /guides/images
 :includes: ./_includes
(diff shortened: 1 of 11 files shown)


holly-cummins:
Even with this change, the link crawler test in dev mode isn't perfectly green, but that's because 

- Some pages don't include the attributes-local
- Some dead image links are pre-existing problems that also affect prod, but the prod crawl sometimes has a thread starvation issue and doesn't crawl the full site on big changes (which I thought I'd fixed, but maybe not). By design, it only crawls affected pages on small changes.

gsmet (approved):

PR #2923, opened 2026-08-24
Restore all config page by avoiding premature conversion of blocks

holly-cummins:
This is kind of subtle, and I'm relying on Claude for diagnosis here. But I *have* verified manually, and added new tests for the all-config page, since it seems like the kind of thing that could go wrong. 

The problem happened on ConfigTableTreeprocessor, which is a Java replacement Marko and I wrote for some of the Ruby Asciidoc plugins. The search input injection code called `caption.getContent()` to read the legend paragraph's HTML. In AsciidoctorJ, this triggers premature conversion of the parent block's content, which causes `Document.convert()` (used by Roq's rendering pipeline) to skip the table  entirely. The table is silently dropped from the output.

The bug requires two conditions:
  1. Table has `.searchable` role (triggers the search input injection path)
  2. Table cells have `[.description]` blocks (triggers the collapsible row
     processing)

This is the combination used by /guides/all-config. I've checked and that combination doesn't seem to happen anywhere else.

To fix it, in ConfigTableTreeprocessor.java , instead of calling `caption.getContent()` and replacing the caption block, we insert the search input as a new sibling block  before the table. This avoids the premature conversion side effect.

I also added the tests I thought I'd added already:

1. AllConfigPageTest.java — Browser test that navigates to /guides/all-config and asserts it returns 200 and has a decent number of config table rows. This  is the test gap that allowed the bug to ship.
2. AsciidocConversionTest.java — Unit-level regression test that reproduces the exact bug: loads a searchable config-reference table with .description  blocks via doc.convert() and asserts the table HTML is present.

### DIFF
--- a/src/main/java/io/quarkus/tools/migration/asciidoc/ConfigTableTreeprocessor.java
+++ b/src/main/java/io/quarkus/tools/migration/asciidoc/ConfigTableTreeprocessor.java
@@ -58,13 +58,17 @@ public class ConfigTableTreeprocessor extends Treeprocessor {
 
             if (table.hasRole("searchable") && table.getParent() instanceof StructuralNode parent) {
                 List<StructuralNode> siblings = parent.getBlocks();
-                int tableIndex = siblings.indexOf(table);
-                if (tableIndex >= 0) {
+				int tableIndex = logicalIndexOf( siblings, table );
+                if (tableIndex > 0) {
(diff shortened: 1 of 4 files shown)


holly-cummins:
I've pushed an improved version of the fix from @marko-bekhta.

marko-bekhta (approved):
🤞🏻

PR #2925, opened 2026-08-24
Image fix for wide monitors and spring page layout change

insectengine:
1) For large monitors, add a max width for images on the /about page so they don't get massive. Sizing based on common 2560x1440 monitor size.

2) refactor /spring page so the roles content is formatted to read as two horizontal bands instead of two verticals for user clarity.


holly-cummins:
I can see when I make the page huge the about images are not so big as on the main site, and I think the spring layout is more readable. Nice!

holly-cummins (approved):

insectengine:
I used to have a 2560x1440 monitor so I never saw this issue.

PR #2927, opened 2026-08-24
Do not deploy only the latest guides in the prod CI, obviously

holly-cummins:
Why is the production build behaving like the preview build, and only showing the latest guides?

Because that's what the build tells it to do. ARGH. 

Turns out `QUARKUS_PROFILE=only-latest-guides` isn't what we want in prod. Copy-paste FTW.


marko-bekhta (approved):
🫣

PR #2929, opened 2026-08-24
Reinstate TOC after Roq migration

holly-cummins:
Guide table of contents got lost in the migration. 

@rolfedh has already implemented TOC support for Roq (https://github.com/quarkiverse/quarkus-roq/pull/776), but there was some discussion about whether it was needed, and the PR hasn't merged. 

What I've done, to not redo work Rolfe has already done, is pull the contents of his PR into the quarkus.io project as a freestanding plugin. I didn't want to vendor the code in and create a fork situation, so instead of set up a build step to copy a fixed commit hash of his PR. When running locally, people can run `roq-toc-plugin/build.sh` if they want a TOC, or just run normally if they don't need it. 

The dependency is optional via a Maven profile that auto-activates when the plugin JAR exists in the local Maven repo. Without it, guides build normally but without a TOC. CI workflows run build.sh before the site build to ensure the plugin is always available in CI. I did have to provide a custom pom and a few other files. 

I also added tests which only run when the plugin is there. It's kind of weak, but it's something.

Obviously, this is a bit tactical, and we hope this support becomes available in Roq. 

As well as the 'building a PR locally' ugliness, there are a few other ugly bits:

- Rolfe's plugin doesn't honour adoc toc levels, so I had to layer in a further hack to get the levels to be 2, rather than 6
- The styling is by eye, because I couldn’t get an exact set of css that could apply to the new html

The original TOC: 

image

The revived version:

image

The styling is fiddly so what I have isn't perfect but I think it's better than no TOC.


rolfedh:
@holly-cummins please let me know if there's anything I can do to help.

rolfedh:
@holly-cummins I pushed two fixes to my fork on `fix-toc`, commit `d30c3175e034a3125e6d092980ca248d58ceee10`. (quarkiverse/quarkus-roq#776)

The failing `RssFeedTest` is my bug, not your wiring. The class-level `@TemplateExtension` registered every static helper in the class, including an `escapeHtml(String)` that replaced the site's own `escapeHtml` and wrote `&nbsp;` into `feed.xml`. Only `toc` and `tocHtml` are registered now, and a test guards that.

The plugin also reads the AsciiDoc `:toclevels:` attribute, defaulting to 2 like Asciidoctor, so you can drop the `sed` patch in `build.sh`. The five guides that set `:toclevels:` explicitly get their own depth, as they did under Jekyll.

Repinning `UPSTREAM_COMMIT` to that hash should turn CI green.

holly-cummins:
**EDIT:** I looked at CI but not issue comments (oops) before writing the below comment. I'll adjust to use Rolfe's new code, thanks! 

The CI failure was kind of subtle. Adding the toc broke the RSS feed (and this is why we have tests!). 

 Qute's @TemplateExtension processor filters out private  methods but NOT package-private ones. So the TOC plugin's static String  escapeHtml(String text) method — which is package-private — gets registered as  a template extension on String.

Claude worked out this chain

1. The TOC plugin's RoqPluginTocTemplateExtension has @TemplateExtension at  class level
2. It contains static String escapeHtml(String text) — package-private, but  Qute registers it anyway
3. This method uses JSoup's Entities.escape() with EscapeMode.base, which converts the Unicode non-breaking space character (U+00A0) to the named HTML entity &nbsp;
4. In the feed template (feed.xml), {=post.content.escapeHtml.raw} calls this method on blog post HTML
5. AsciiDoc-rendered blog content contains literal U+00A0 characters (common in tables, callouts)
6. The JSoup escaper converts them to &nbsp; — valid in HTML, but not a defined entity in XML
7. The XML parser in RssFeedTest dies, correctly: "The entity "nbsp" was referenced, but not declared."

holly-cummins:
I've picked up @rolfedh's latest, dropped my hack-patches, and things are almost working nicely:

image

However, I'm not picking up the deeper nesting for some reason, and I lost numbering. Logging should have 4 levels: 

image

I won't push because I think the previous version was arguably slightly better in terms of the output (at the expense of being uglier in the code). Either way we'd want to fix the levels to read what's in the doc, but I *think* it would be ok to merge the previous one and iterate.

gsmet (approved):

PR #2930, opened 2026-08-25
Fix missing version options on single guide views after Roq migration

marko-bekhta:
Apparently, the guides never "matched" in other versions, so everything but "current" was filtered out. And that trailing `/` didn't help either, sicne it's not in the data files 🙈 😖

### DIFF
--- a/templates/layouts/guides.html
+++ b/templates/layouts/guides.html
@@ -14,7 +14,7 @@ layout: base
  !}
 {#let page_filename=page.sourcePath.replace('_guides/', '').replace('_versions/main/guides/', '')}
 {#let relations=cdi:versioned.get(docversion_index).index.relations}
-{#let guide_url=page.url.path.replaceAll('^/version/[^/]+(/.*)', '$1')}
+{#let guide_url=page.url.path.replaceAll('^/version/[^/]+(/.*)', '$1').replaceAll('/$', '')}
 
 <section class="full-width-version-bg flexfilterbar guides">
   <div class="guideflexcontainer">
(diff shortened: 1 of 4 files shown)


templates/layouts/guides.html
```diff
@@ -37,13 +37,9 @@
...
         {/for}
         {#let data_source=version.replace('.', '-')}
-        {#for type in cdi:versioned.get(data_source).index.quarkus.types.orEmpty}
-          {#for guide in type.get(1).orEmpty}
-            {#if guide.url == guide_url}
-              {=_m.assign('guide_present_in_version', true).raw}
-            {/if}
-          {/for}
-        {/for}
+        {#if cdi:versioned.get(data_source).index.containsGuide(guide_url)}
```
holly-cummins, line 40: What's the reason for skipping the for iteration? I'm sure it's correct, I just want to understand the code better.

marko-bekhta: ah, yeah, we do need to run through the list of guides to find one, it just that these for loops now moved inside the `containsGuide` helper instead 🙈 

Comparing both -- the Java version seems sooo much more verbose than the in-template original 🙈 

but since we have it in a couple of places I thought that helper would be better to share the check ... 
Or, alternatively, we just have a small include template for this for loop 🤔

holly-cummins: Got it, thanks! Let's leave as you have it and we can refine if it annoys us (but I think your move to sharing it was sensible.)

holly-cummins (approved):

PR #2931, opened 2026-08-25
Fix older guide versions navigation to not trigger downloads from index page

holly-cummins:
For older guides versions (anything with a dot), on the live site, but not in previews, and not in dev mode, trying to view a page by clicking on it on the index page does a download instead of a navigation. The root cause seems to be https://github.com/quarkiverse/quarkus-web-bundler/issues/473, which is muddling dots-in-folders with dots-extensions. Happily, it can be worked around with some template updates. 

I got annoyed at preview and CI tests being 'blind' for versions, so I've updated CI to detect infra changes and build all versions. That allows guides tests to run against older versions. I also, in these cases, updated the preview to include a handful of older versions, but not all of them. (The index always includes them all, which is a but confusing, but we can live with that for now.)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -91,9 +91,33 @@ jobs:
         if: steps.detect.outputs.roq == 'true'
         run: ./roq-plugin-toc/build.sh
 
+      - name: Detect infra changes
+        id: infra
+        run: |
+          if [ -n "${{ github.event.pull_request.base.sha }}" ]; then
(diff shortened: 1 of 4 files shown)


holly-cummins:
None of the releases that are published on the preview show in the drop down, which is a bit annoying. But since this fixes a user-facing issue I'd propose we merge and then I iterate on the CI preview tweaking (which is no worse than the current state)

marko-bekhta (approved):
Let's merge it and see in prod ! 🙂

PR #2934, opened 2026-08-25
Fix fragment ids after Roq migration

holly-cummins:
@ppalaga spotted that some of our fragment links regressed after migration. I checked, and about 50 had regressed. This should have been caught by https://github.com/quarkusio/quarkusio.github.io/pull/2713, but that PR is still failing because of _other_ dead fragment links. 

Asciidoctor defaults to idprefix='_' and idseparator='_', but jekyll-asciidoc overrode these to '' and '-', producing IDs like 'my-section' instead of '_my_section'. This can almost be controlled by application-properties, but SmallRye Config rejects empty map values so idprefix cannot be set to an empty value via application.properties. Instead, I set both attributes at the document level instead, using the global attributes. The faq.adoc doesn't inherit the attributes so in that document I edited the frontmatter directly.

I've tested this fix with #2713 and the links are back to passing.

### DIFF
--- a/content/faq.adoc
+++ b/content/faq.adoc
@@ -3,6 +3,8 @@ layout: base-fullwidthcontent
 title: FAQ
 subtitle: Get answers to some of your common Quarkus questions.
 ---
+:idprefix:
+:idseparator: -
 
 == What is your license?
 
--- a/content/guides/_attributes-local.adoc
+++ b/content/guides/_attributes-local.adoc
@@ -1,3 +1,5 @@
+:idprefix:
+:idseparator: -
 // tag::xref-attributes[]
 :doc-examples: ./_examples
 :doc-guides: .


holly-cummins:
https://quarkus-website-pr-2934-preview.surge.sh/faq/#what-is-a-quarkus-extension now works.

gsmet (approved):

gsmet:
TBH, I'm wondering if it's something that should be adjusted at the Roq level. The default settings are not really ideal if you ask me.

holly-cummins:
> TBH, I'm wondering if it's something that should be adjusted at the Roq level. The default settings are not really ideal if you ask me.

I wondered that too. Jekyll overrides the Asciidoc defaults (which are weird), and Roq does not. But maybe it's reasonable to assume most people using Asciidoc would want the Asciidoc defaults? 

What I have done is raise https://github.com/quarkiverse/quarkus-roq/pull/1175, which makes it easier to adjust the settings in Roq with global config. (Because the config is changing a something to a nothing, it's not really compatible with smallrye config without hackery.)

gsmet:
I don't know, I think having better defaults would be a win. But that's for @ia3andy to decide.

PR #2936, opened 2026-08-25
Give guides the TOC depth their toclevels asks for

rolfedh:
## TLDR

Guides now get the table of contents depth their AsciiDoc `:toclevels:` asks for, as they did before the Roq migration. This repins the TOC plugin to a commit that reads that attribute and removes both `sed` patches from `roq-plugin-toc/build.sh`. Section numbering is unchanged.

Follow-up to #2929, which reinstated the TOC at a fixed depth of three heading levels as a temporary measure.

## What changed

- `roq-plugin-toc/build.sh` is pinned to `94dc2ccb` on the `fix-toc` branch of https://github.com/quarkiverse/quarkus-roq/pull/776. That commit reads each document's `toclevels`, so the patch that capped every guide at three heading levels is gone, and it registers only its two template extensions, so the patch that made the escaping helpers private is gone too.
- `UPSTREAM_REPO` and `UPSTREAM_COMMIT` can now be overridden from the environment, so you can build the plugin from a local checkout while iterating.
- `GuideTocTest` pins the depth of six pages and the numbering of four.

## Evidence

Depth is the maximum `data-level` in `nav.roq-toc`, counting heading tags from `h1` = 0. The last column is Asciidoctor's own `:toc:` outline for the same source, which is what the Jekyll site's `tocify_asciidoc` filter rendered.

| Guide | `:toclevels:` | Before | After | Asciidoctor |
|---|---|---|---|---|
| logging | 4 | 2 | 4 | 4 |
| qute-reference | 3 | 2 | 3 | 3 |
| getting-started | none | 2 | 2 | 2 |
| cdi-integration | 2 | 2 | 2 | 2 |
| security-oidc-code-flow-authentication | 4 | 2 | 3 | 3 |

The OIDC guide asks for four section levels but has no section deeper than `h4`, so three is the correct result for it.

Section numbering is unchanged and was never affected: Asciidoctor puts the number in the heading text, and only for guides that set `:sectnums:` or `:numbered:`. qute-reference, getting-started and cdi-integration are numbered before and after; logging and the OIDC guide set neither attribute and were unnumbered on the old site too.

## How this was verified

- Full batch build (`QUARKUS_ROQ_GENERATOR_BATCH=true mvn package quarkus:run`), 5341 pages, then the same check in dev mode. Both give the results above.
- `mvn test -Dtest.url=...` against the built site: `GuideTocTest` 18 tests and `RssFeedTest` 5 tests pass. `RssFeedTest` is the one that failed in #2929 before the plugin stopped registering its escaping helper as a Qute extension.
- 44 plugin unit tests pass in `build.sh`.

## Not fixed here

Guides that declare an anchor before their title, such as logging, still show `guides/logging.adoc` as the page heading. That is https://github.com/quarkiverse/quarkus-roq/issues/1136, fixed upstream in yupiik/tools-maven-plugin#92 but not yet in the parser version Roq pins. It is unrelated to the TOC.

### DIFF
--- a/roq-plugin-toc/build.sh
+++ b/roq-plugin-toc/build.sh
@@ -3,8 +3,15 @@
 # Builds the Roq TOC plugin from rolfedh/quarkus-roq PR #776 at a pinned commit,
 # using standalone POMs that decouple it from the Roq monorepo build.
 #
+# The pinned commit reads each guide's AsciiDoc :toclevels: attribute and registers only
+# its two template extensions, so the sources are built unmodified.
+#
 # Usage:
 #   ./roq-plugin-toc/build.sh          # fetch, build, and install to local Maven repo
+#
+# Set UPSTREAM_REPO and UPSTREAM_COMMIT to build from a local checkout instead, e.g.
+#   UPSTREAM_REPO=file:///path/to/quarkus-roq \
+#   UPSTREAM_COMMIT=$(git -C /path/to/quarkus-roq rev-parse fix-toc) ./roq-plugin-toc/build.sh
 #
 # The plugin is then available as:
 #   io.quarkiverse.roq:quarkus-roq-plugin-toc:1.0.0-SNAPSHOT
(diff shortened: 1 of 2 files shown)


rolfedh:
@holly-cummins FYI

gsmet:
@holly-cummins I let you drive this one home.

PR #2938, opened 2026-08-26
Fix the All configuration options page

gsmet:
See comments in #2900 . Make the page consistent with the rest of the guides.

Thanks to @mkouba for the hidden fragment trick that prevents the duplication.

/cc @insectengine @holly-cummins


holly-cummins:
https://quarkus-website-pr-2938-preview.surge.sh/guides/all-config/ is looking as it should.

holly-cummins (approved):

PR #2943, opened 2026-08-26
Restore the title anchors

### LINKED ISSUE #2942: Guides - headers do not contain anchors after migration to Roq
**Describe the bug**

Previously all headers contained an anchor, something like this:
```html
<h2 id="_protocol_support"><a class="anchor" href="#_protocol_support"></a>Protocol Support</h2>
```
Wich is useful because you can click on the anchor and the URL is changed and can be easily copied and used in a discussion for example...

**Which pages**

https://quarkus.io/guides/

gsmet:
Fixes #2942

Not sure if the default should be changed in Roq. /cc @ia3andy @holly-cummins

### DIFF
--- a/web/_asciidoc.scss
+++ b/web/_asciidoc.scss
@@ -3,7 +3,7 @@
 table tr th, table tr td {padding: .5625em .625em;}
 
 #content h1>a.anchor,h2>a.anchor,h3>a.anchor,#toctitle>a.anchor,.sidebarblock>.content>.title>a.anchor,h4>a.anchor,h5>a.anchor,h6>a.anchor{position:absolute;z-index:1001;width:1.5ex;margin-left:-1.5ex;display:block;text-decoration:none!important;visibility:hidden;text-align:center;font-weight:400}
-#content h1>a.anchor::before,h2>a.anchor::before,h3>a.anchor::before,#toctitle>a.anchor::before,.sidebarblock>.content>.title>a.anchor::before,h4>a.anchor::before,h5>a.anchor::before,h6>a.anchor::before{content:"\00A7";font-size:.85em;display:block;padding-top:.1em}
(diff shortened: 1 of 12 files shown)


web/_asciidoc.scss
```diff
@@ -3,7 +3,7 @@
 table tr th, table tr td {padding: .5625em .625em;}
 
 #content h1>a.anchor,h2>a.anchor,h3>a.anchor,#toctitle>a.anchor,.sidebarblock>.content>.title>a.anchor,h4>a.anchor,h5>a.anchor,h6>a.anchor{position:absolute;z-index:1001;width:1.5ex;margin-left:-1.5ex;display:block;text-decoration:none!important;visibility:hidden;text-align:center;font-weight:400}
-#content h1>a.anchor::before,h2>a.anchor::before,h3>a.anchor::before,#toctitle>a.anchor::before,.sidebarblock>.content>.title>a.anchor::before,h4>a.anchor::before,h5>a.anchor::before,h6>a.anchor::before{content:"\00A7";font-size:.85em;display:block;padding-top:.1em}
+#content h1>a.anchor::before,h2>a.anchor::before,h3>a.anchor::before,#toctitle>a.anchor::before,.sidebarblock>.content>.title>a.anchor::before,h4>a.anchor::before,h5>a.anchor::before,h6>a.anchor::before{content:"\00A7";font-size:.85em;display:block;padding-top:.4em}
```
holly-cummins, line 6: Are these styling changes part of this fix?

gsmet: Yes. The anchors are not at the right place if not.

holly-cummins:
> Not sure if the default should be changed in Roq. /cc @ia3andy @holly-cummins

I've had a look, and the quarkus config always set the section anchors property. (That is, in other generators, the defaults are the same as what Roq uses.) The problem happened because just that Roq couldn't really support the global section anchors property because of the limitation that smallrye can't handle empty strings. That meant the property got lost in conversion. So I suggest we add the missing support, but do not change the defaults.

gsmet:
I'm merging this as this is very annoying and I will let you figure out what you want to do for the future :).

holly-cummins:
I've raised https://github.com/quarkiverse/quarkus-roq/pull/1182 for Roq support of this mirroring what the Jekyll asciidoctor config can do.

PR #2944, opened 2026-08-26
Patch Roq adoc tag processing to fix empty includes

### LINKED ISSUE #2939: Java source includes don't work anymore
See https://quarkus.io/version/main/guides/telemetry-micrometer-tutorial/#_add_a_counter .

Claude is saying that Roq might only support adoc includes. Which might or might not be true. It might also be a path issue, I don't know.

Also it's annoying that we don't get an error when the include is not optional. We should have one.

It needs a fix.

/cc @holly-cummins @ia3andy

holly-cummins:
Resolves https://github.com/quarkusio/quarkusio.github.io/issues/2939

The correct fix for this is obviously to do https://github.com/quarkiverse/quarkus-roq/issues/1178, but this is pretty severe, so I think a patch is warranted. I experimented with Quarkus Shim, but it couldn't patch 'deep' enough to get over classloader isolation in this case, so I've done a manual update of the jar. We'll want to remove that as soon as we can. 

I've also added a test to make sure we don't have this kind of thing again (it will also protect against other include problems, like missing files). The test will fail locally unless people run the patch script, but I think that's ok as a very short term thing.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -91,6 +91,13 @@ jobs:
         if: steps.detect.outputs.roq == 'true'
         run: ./roq-plugin-toc/build.sh
 
+      # TODO: Remove once https://github.com/quarkiverse/quarkus-roq/issues/1178 is fixed and Roq is upgraded
+      - name: Patch Roq AsciiDoc tag handling
+        if: steps.detect.outputs.roq == 'true'
+        run: |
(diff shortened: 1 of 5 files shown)


holly-cummins:
See also https://github.com/quarkiverse/quarkus-roq/pull/1180 for a bit more explanation of the fix.

gsmet (approved):

Issue #1178, opened 2021-10-18
Add How to create Quarkus Command Mode applications - Mastertheboss to Publications


jclingan:
```
- id: HowtocreateQuarkusCommandModeapplicationsMastertheboss-2021-10-18
  type: article 
  title: "How to create Quarkus Command Mode applications - Mastertheboss"
  date: "2021-10-18"
  author: "Francesco Marchioni"
  note: "In this tutorial we will learn how to create standalone Quarkus applications with a bare simple main entry point."
  url: "http://www.mastertheboss.com/soa-cloud/quarkus/how-to-create-quarkus-command-mode-applications/?utm_source=dlvr.it&utm_medium=twitter&utm_campaign=how-to-create-quarkus-command-mode-applications"
  urldate: "2021-10-18"

```

Issue #1158, opened 2021-10-06
Add Extension Codestart guide to the list of guides


ia3andy:
Add https://quarkus.io/version/main/guides/extension-codestart to the list of guides

ia3andy:
cc @maxandersen I am not sure how to do that?

Issue #776, opened 2020-10-29
Antonio Goncalves book


maxandersen:

insectengine:
What is a good title for this Episode?
"New Books about Quarkus"

PR #2955, opened 2026-08-27
Replace the temporary TOC plugin build with the released Roq plugin

### LINKED ISSUE #2954: Replace the temporary TOC plugin build with the released Roq plugin
The guide table of contents comes from the Roq TOC plugin (quarkiverse/quarkus-roq#776). #2929 wired it in temporarily: `roq-plugin-toc/build.sh` clones a pinned commit from a personal fork on every CI run and installs the plugin into the local Maven repository, and a Maven profile picks it up when the jar is there.

Replace all of that with one dependency on `io.quarkiverse.roq:quarkus-roq-plugin-toc` and remove:

- `roq-plugin-toc/` (`build.sh` and the three standalone POMs)
- the `toc` profile in `pom.xml`
- the "Build TOC plugin" steps in `.github/workflows/build.yml` and `deploy.yml`
- the "skip if the plugin is absent" assumption in `GuideTocTest`

Status (4 September): #776 merged on 31 August, and Roq 2.1.8 and 2.1.9 contain the plugin. Holly merged the 2.1.9 upgrade (#2972) and the jar-patch removal (#2983). #2955 makes this change and is ready for review.

cc @gsmet @holly-cummins

rolfedh:
Fixes #2954.

Removes the temporary wiring from #2929 and depends on the released plugin instead: `roq-plugin-toc/` and its `build.sh`, the `toc` Maven profile, the "Build TOC plugin" CI steps, and the "skip if the plugin is absent" assumption in `GuideTocTest` are gone, and `io.quarkiverse.roq:quarkus-roq-plugin-toc` is a normal dependency using the existing `roq.version` property.

quarkiverse/quarkus-roq#776 was merged on 31 August and is in Roq 2.1.8 and 2.1.9. This PR stays draft, and CI stays red, until #2972 (the upgrade to 2.1.9) merges, because main still pins 2.1.7 and there is no 2.1.7 `quarkus-roq-plugin-toc` artifact. After #2972 merges I'll rebase this PR and mark it ready. Per the discussion below, this PR stays TOC-only.


holly-cummins:
When we move to the next Roq release, it's likely we will have CI failures where local workarounds fight with released fixes. That means when this PR merges/runs, there would be test failures. We can either group everything, or merge one at a time and accept a bit of red while we unpeel the workarounds

### Will need reverting before/with update
- https://github.com/quarkusio/quarkusio.github.io/pull/2947 (assuming https://github.com/quarkiverse/quarkus-roq/pull/1183 merges)

### Not sure - definitely revert, but CI may be ok in the interim
- https://github.com/quarkusio/quarkusio.github.io/pull/2946 (assuming https://github.com/quarkiverse/quarkus-roq/pull/1158 merges)
- https://github.com/quarkusio/quarkusio.github.io/pull/2919 (also a https://github.com/quarkiverse/quarkus-roq/pull/1158 one)
- https://github.com/quarkusio/quarkusio.github.io/pull/2931 - webbundler fix has merged. Should only revert the template change, not the test strenghtening

### Harmless but should be reverted after updating
These should be reverted because they're clutter, but I don't think they will break anything in the short term

- The ToC change, reverted by this PR :)  
- https://github.com/quarkusio/quarkusio.github.io/pull/2951
- Some redirect exclusions in the LinkCrawler test
- https://github.com/quarkusio/quarkusio.github.io/pull/2961 and https://github.com/quarkusio/quarkusio.github.io/pull/2943, assuming https://github.com/quarkiverse/quarkus-roq/pull/1175 is in the release

rolfedh:
Agreed, grouping is better. When 2.1.8 is out I'd turn this PR into the upgrade PR: bump the Roq versions, remove the "Patch Roq for pre-2.1.8 workarounds" step and `patches/`, and add the TOC dependency if #776 has landed. Both patched fixes are on Roq main (the `;` separator fix for #1178 and the yupiik bump in #1184), and diffing the patched classes against main shows main already contains everything they add, so keeping the patches would overwrite newer code. #2931 has to stay: web-bundler#474 merged, but the latest web-bundler release is still 2.3.3, which Roq main pins. #2947, #2946 and #2919 go in whichever upgrade first ships quarkus-roq#1183 and #1158.

On `roq.version`: the patch scripts never read it (their grep doesn't match, so the hard-coded fallback runs) and #2953 already added the property on `main`, so this PR only needs a rebase. I rehearsed the upgrade locally against Roq main plus the TOC branch with the patches removed: all 271 smoke tests pass. Inventory to follow on #2954.

*(Edited: my first version promised to go back to literal versions; that was based on a wrong reading of the scripts.)*

holly-cummins:
My personal preference would be for smaller PRs, especially if #2947 is the only actually-breaking-one. I'm also not sure if the Roq-side PR for that will be in the next release or not.

holly-cummins:
Also, https://github.com/quarkusio/quarkusio.github.io/pull/2928 should merge before any upgrade (and the upgrade PR should be rebased onto it).

holly-cummins:
#2928 is now in, but https://github.com/quarkusio/quarkusio.github.io/pull/2713 should also merge before update (and this rebase onto it).

rolfedh:
Agreed, @holly-cummins. Let's use smaller PRs. Here is how I propose to split the work:

- This PR (#2955) only replaces the TOC plugin build with the released plugin. It waits until a Roq release contains quarkiverse/quarkus-roq#776.
- A separate, minimal PR upgrades Roq to 2.1.8. It sets `roq.version` to 2.1.8, deletes the "Patch Roq for pre-2.1.8 workarounds" step and the `patches/` directory (the step is already labelled "remove with 2.1.8"), and reverts #2947 only if the release contains quarkiverse/quarkus-roq#1183. I will open that PR after 2.1.8 is released and #2713 is merged, unless you prefer to open it yourself, since you wrote most of the workarounds. Please tell me which you prefer.
- Each remaining workaround (#2931, #2946, #2919 and the others) gets its own PR when the Roq or web-bundler fix it waits for is released.
- After #2713 merges, I will rebase this PR onto `main`.

rolfedh:
@holly-cummins Roq 2.1.8 is out (31 August, on Maven Central). I verified the tag: it contains quarkiverse/quarkus-roq#776 (the TOC plugin) and quarkiverse/quarkus-roq#1158, and it pins web-bundler 2.3.4, which contains the quarkiverse/quarkus-web-bundler#474 fix. That means the upgrade unlocks everything except the #2947 revert (quarkiverse/quarkus-roq#1183 is still open), and this PR can go right after it. I'm ready to open the upgrade PR as soon as #2713 merges — or you open it, whichever you prefer.

rolfedh:
@holly-cummins here's the sequence I propose, now that 2.1.8 has everything we need:

- You merge #2713. I start nothing before that, as agreed.
- I rebase this PR to one clean commit on main.
- I open a minimal upgrade PR that bumps `roq.version` to 2.1.8 and deletes the three "Patch Roq for pre-2.1.8 workarounds" steps and the `patches/` directory. Deleting `patches/` also removes the yupiik override, which now lives in `patches/pom.xml`. I leave #2947 in place, because quarkiverse/quarkus-roq#1183 is still open.
- After you merge the upgrade PR, I mark this PR ready and ask you to review it.
- I then open two small revert PRs: one for #2931 and one for #2946/#2919.

If you'd rather open the upgrade PR or the reverts yourself, tell me and I'll stick to the TOC work.

holly-cummins:
Ok, I think we're ready to proceed with the plan. I'll do a just-2.1.8 PR so we can see if anything breaks. Actually, I think the #2947 fix needs to be reverted. So maybe I'll do those two first and then @rolfedh can pick up the next step of his plan. I'll do just 2.1.8 in CI first to confirm I'm right about the failures.

rolfedh:
Sounds good, @holly-cummins  — thanks for taking the upgrade PR. One thing that may save you a CI run: I don't think #2947 can be reverted with 2.1.8. Reverting it restores links like `/blog/vscode-quarkus-1.2.0/`, but Roq still generates `/blog/vscode-quarkus-1-2-0/`. The dotted URLs come back only with quarkiverse/quarkus-roq#1183, which is still open — `TemplateLink.java` is identical at the 2.1.7 and 2.1.8 tags, and no slug or link generation file changed between them. Happy to be proved wrong by CI.

holly-cummins:
> Sounds good, @holly-cummins — thanks for taking the upgrade PR. One thing that may save you a CI run: I don't think #2947 can be reverted with 2.1.8. Reverting it restores links like `/blog/vscode-quarkus-1.2.0/`, but Roq still generates `/blog/vscode-quarkus-1-2-0/`. The dotted URLs come back only with [quarkiverse/quarkus-roq#1183](https://github.com/quarkiverse/quarkus-roq/pull/1183), which is still open — `TemplateLink.java` is identical at the 2.1.7 and 2.1.8 tags, and no slug or link generation file changed between them. Happy to be proved wrong by CI.

No, I agree. So there's nothing breaking in Roq 2.1.9, and we can roll back the other workarounds on a schedule and in an order that suits us.

rolfedh:
This PR waits for #2983, and the red CI here comes from the patch step, not from this change. Once #2983 merges I'll rebase and mark this ready.

The "Patch Roq" step still on main replaces `pom.xml` with `patches/pom.xml`, an old copy without the `quarkus-roq-plugin-toc` dependency, so this build had no TOC plugin at all and `GuideTocTest` failed on every case. Every other test passed.

rolfedh:
@holly-cummins #2955 is rebased onto main (one commit, 9 files) and CI is green now that main pins 2.1.9. I've marked it ready — could you review? It only swaps the temporary plugin build for the released dependency; nothing else changed.

holly-cummins (approved):
Excellent to get rid of the workaround. I was faintly concerned by the dropping of the empty toc test, but I guess if the new behaviour annoys us, we can fix.

rolfedh:
The empty-TOC case is still covered. A guide without section headings (security-identity-providers, for example) gets an empty `<div class="toc"></div>` — the plugin returns an empty string — and the existing rule `body.guides div.guide .toc:empty { display: none }` hides it; I checked the live page and the compiled CSS. The dropped test never exercised that case: it was hard-wired to `/guides/getting-started`, which has headings, so with the plugin installed it asserted nothing. If you'd rather have an explicit test, one that loads a heading-less guide and checks the wrapper is hidden is small — say the word.

holly-cummins:
Ah, even better, thanks!

PR #2957, opened 2026-08-27
Fix favicon path

gsmet:


holly-cummins:
Will update the converter as well.

PR #2959, opened 2026-08-28
Do not use latest yupiik in non-CI builds to avoid incompatible binary

holly-cummins:
This should all go away soon, but in the interim, normal local builds should work.

I've checked the prevew, and guide titles are still correct, I *think*. Only some guides had the issue so I checked a few and see 

image


gastaldi (approved):
LGTM

PR #2961, opened 2026-08-28
Extend idprefix/idseparator fix to blog posts and versioned guides

holly-cummins:
PR #2934 set :idprefix:/:idseparator: - at the document level to restore the jekyll-asciidoc defaults lost in the Roq migration, but only for the FAQ and the live guides' _attributes-local.adoc. While trying to get #2713 in, I spotted the gap. This extends the approach to the versioned guides and also some blog posts.

As before, https://github.com/quarkiverse/quarkus-roq/pull/1175 is the more compact fix, but not yet available.

Even with this, #2713 isn't green, but it gets it closer.

### DIFF
--- a/content/posts/2019-09-23-quarkus-developer-joy-for-vs-code.adoc
+++ b/content/posts/2019-09-23-quarkus-developer-joy-for-vs-code.adoc
@@ -6,6 +6,8 @@ date: 2019-09-23
 synopsis: Showcasing the new Quarkus VS Code extension.
 author: dakwon
 ---
+:idprefix:
+:idseparator: -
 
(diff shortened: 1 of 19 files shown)


gastaldi (approved):
LGTM, but can we add that to a common template (`posts`) to reuse?

holly-cummins:
> LGTM, but can we add that to a common template (`posts`) to reuse?

I'm *hoping* that once https://github.com/quarkiverse/quarkus-roq/pull/1175 merges, all this becomes unnecessary (and we can rip it back out). I agree, otherwise it should be a template if that doesn't work out.

PR #2972, opened 2026-09-01
Bump Roq

holly-cummins:
See  https://github.com/quarkusio/quarkusio.github.io/pull/2955

I'm expecting this to fail with some dead links, but will do a CI run to be sure.

Confirmed failing, but not in the way I was expecting: 27 failures of the form 

```
404 http://localhost:8090/https:/quarkus.io/guides/deploying-to-kubernetes/ ← http://localhost:8090/guides/init-tasks/
```

(single slash instead of double after https)

https://github.com/quarkusio/quarkusio.github.io/pull/2947/changes didn't seem to cause problems.

Edit: I've raised a Roq PR with a fix, waiting for a new Roq release.

### DIFF
--- a/patches/apply-roq-include-fix.sh
+++ b/patches/apply-roq-include-fix.sh
@@ -14,7 +14,7 @@ PROJECT_DIR="$(cd "$SCRIPT_DIR/.." && pwd)"
 # Extract the Roq version from pom.xml
 ROQ_VERSION=$(grep -m1 -oP '(?<=quarkus-roq-plugin-asciidoc-jruby</artifactId>\s{0,99}<version>)[^<]+' "$PROJECT_DIR/pom.xml" \
   || grep -oP '(?<=quarkus-roq</artifactId>\s{0,99}<version>)[^<]+' "$PROJECT_DIR/pom.xml" \
-  || echo "2.1.8")
+  || echo "2.1.9")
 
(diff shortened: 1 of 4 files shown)


holly-cummins:
CI is now green. The change that might have caused breaking URL (re-)changes didn't make it into the release, sadly, so nothing in the 2.1.9 Roq release breaks quarkus.io externals. I think we can merge this and then move on to @rolfedh's PRs.

rolfedh (commented):
Thanks for confirming #2947 — this looks good to merge. As soon as it lands I'll open a follow-up PR that deletes the three "Patch Roq for pre-2.1.8 workarounds" steps and `patches/`, unless you'd rather delete them here.

They no longer do what they were built for. The include script's fallback now says 2.1.9, and because its version grep never matches, that fallback runs: the green run's log shows `Patching .../quarkus-roq-plugin-asciidoc-jruby/2.1.9/...jar`, so CI replaces the 2.1.9 `AsciidocJInclude` with the 2.1.7-era workaround copy. The header script still says 2.1.7, so it downloads and patches a jar the build never loads.

Removing them is safe: 2.1.9 contains both upstream fixes, and my 27 August rehearsal passed all 271 smoke tests with the patches off.

rolfedh (approved):
Approved. Nothing changed since yesterday's review: the bump to 2.1.9 is correct, the plugin artifacts are on Maven Central, and CI is green. I'll open the patch-removal PR as soon as this merges.

PR #2976, opened 2026-09-02
Make blogging process clearer 

holly-cummins:
People sometimes ask on zulip if they're allowed to write blogs, which makes me think we don't do a good enough job of making it obvious they are. We also don't have any google-find-able instructions for writing a blog. I've added something on the main blog page to solicit contributions and done a bit of cross-linking in the source markdown docs. 

image

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,3 +4,5 @@ Contributions are welcome, please submit pull requests for the `main` branch.
 
 **Important:** the guides are maintained in the main Quarkus repository and pull requests should be submitted there:
 https://github.com/quarkusio/quarkus/tree/main/docs/src/main/asciidoc.
+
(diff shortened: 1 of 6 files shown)


gsmet (commented):
I think if you want to do that, we will need guidelines of what we accept and what we don't.

holly-cummins:
> I think if you want to do that, we will need guidelines of what we accept and what we don't.

Possibly, but would it be better to iterate? At the moment, we do accept community contributions, and we don't have guidelines of what we accept and what we don't. This change starts to provide basic guidelines of what we accept: "Quarkus stories, expertise, and ecosystem news".  We *could* be more expansive somewhere else, but I don't think that should block writing down our current policy. This PR isn't changing policy, it's making our current policy more visible and trying to reduce traffic on zulip for this FAQ.

gsmet:
> Possibly, but would it be better to iterate?

Well, I will let you filter and say no when we don't want the contributions, then :)

PR #2977, opened 2026-09-02
Remove Jekyll from CI and remaining readme references

holly-cummins:
I think we're ready to get rid of Jekyll from the actions workflows. Note that the deploy change is risky because it doesn't get tested until we actually do a deploy. 

I've also tidied up some lingering references in the readme and skills.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ These instructions will get you a copy of the Quarkus.io website up and running 
 > Listening on: http://0.0.0.0:8042
 > ```
 
-5. Now browse to [http://localhost:4000](http://localhost:4000).
+5. Now browse to [http://localhost:8042](http://localhost:8042).
 
 
 ### Deploying to GitHub Pages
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -42,49 +42,8 @@ These instructions will get you a copy of the Quarkus.io website up and running
 > Listening on: http://0.0.0.0:8042
 > ```
 
- If an error occurs mentioning a name conflict, try:
-```sh
-docker compose up --force-recreate
-```
-
 5. Now browse to [http://localhost:4000](http://localhost:4000).
```
marko-bekhta: probably ? 
```suggestion
5. Now browse to [http://localhost:8042](http://localhost:8042).
```

based on the listening port above 🙈

holly-cummins: Oh man, I search and replaced for 4000, but apparently not well enough. Thanks!

marko-bekhta (approved):
🎉

PR #2983, opened 2026-09-03
Remove the pre-2.1.8 Roq jar patches

rolfedh:
Removes the three "Patch Roq for pre-2.1.8 workarounds" CI steps and the `patches/` directory. Follow-up to #2972, as discussed there.

Roq 2.1.9 contains both fixes the patches carried: the include fix (quarkiverse/quarkus-roq 9419f3fe, for #2944) and the yupiik asciidoc-java 1.2.16 bump (quarkiverse/quarkus-roq#1184, for #2951). Since the upgrade, the include script was patching the 2.1.9 `asciidoc-jruby` jar with the 2.1.7-era `AsciidocJInclude` class, and the header script was patching a 2.1.7 jar the build no longer loads. `patches/pom.xml`, the copy of the root pom that CI swapped in to carry the yupiik override, goes with the directory; 2.1.9 manages the same version itself.

The rehearsal in #2954 built the site on Roq 2.1.8-to-be without these patches and passed all 271 smoke tests.

Please merge this before #2955: the patch step's pom swap discards #2955's dependency change, so #2955 stays red until this lands.


rolfedh:
@holly-cummins could you review and merge this one first? It is the follow-up I promised on #2972, CI is green, and #2955 stays red until it lands. (The first push missed the two workflow edits, hence the earlier red run — fixed.)

The order from here:

1. You merge this PR.
2. I rebase #2955, wait for green CI and mark it ready; you review and merge it, which closes #2954.
3. I open one small revert PR for #2931 and one for #2946/#2919 — unless you'd rather do those yourself.
4. #2947 waits for quarkiverse/quarkus-roq#1183 and a Roq release that contains it.

holly-cummins (approved):
So much deletion, love to see it.

holly-cummins:
For step 3, I think each of those PRs has a mix of changes we want to keep, and changes we want to revert. I should have structured the commits more carefully, with hindsight. But it probably makes sense for me to handle them, since I remember what's supoosed to stay and what should be reverted.

rolfedh:
Thanks for the quick merge. Agreed — I'll leave both reverts to you and review them when they appear, since you know which parts of #2931 and #2946/#2919 should stay. One note for #2946/#2919: CI doesn't cover dev mode, which is what #2919 fixed, so a quick `quarkus dev` cross-reference check is worth doing there. #2955 is rebased and will be ready for review as soon as its CI is green.

PR #2984, opened 2026-09-04
Move away from deprecated darken, lighten, and import constructs in css

holly-cummins:
The sass processor used by Roq is a bit grumpier about deprecation than the one used by Jekyll, so we're getting deprecation warnings about `@import` in the logs. I've used Bob to fix things up, but it's an invasive change. All looks well to me running locally but I don't have a deep knowledge of sass to be able to judge if Bob made good design choices. I passed Bob's work through Claude, which made a few further revisions.


gsmet:
I will let @insectengine review this one.

insectengine (commented):
Sorry for the delay.

insectengine (approved):

PR #2986, opened 2026-09-04
Let Qute type-check page.tocHtml now that the TOC plugin is a real dependency

rolfedh:
Removes `io.quarkiverse.roq.frontmatter.runtime.model.Page.tocHtml` from `quarkus.qute.type-check-excludes`. Follow-up to #2955.

The exclusion dates from when quarkus.io built the TOC plugin from a fork and installed it locally, so Qute could not rely on seeing its template extension at build time. Since #2955 the plugin is a normal dependency, and Qute resolves `page.tocHtml` through the extension method. Verified on main: with the exclusion removed, `mvn package -DskipTests` passes, and a deliberately misspelt `page.tocHtmlX` fails the build with "Property/method [tocHtmlX] not found on class [Page] nor handled by an extension method" — so the type check now guards the expression instead of being bypassed.

The other excludes (`paginator`, `tags`, `tagsCount`, `DocumentPage.*`) are unrelated and stay.


rolfedh:
@holly-cummins could you review this one-liner when convenient? It drops the `Page.tocHtml` type-check exclusion that only existed for the fork-built plugin; CI is green, and I verified that a misspelt `page.tocHtmlX` now fails the build, so the check is real again.

holly-cummins (approved):
Oh good spot

PR #2987, opened 2026-09-04
Revert relfile prefixes

holly-cummins:
Reverts https://github.com/quarkusio/quarkusio.github.io/pull/2919 now that it is no longer needed, with the Roq update.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -56,8 +56,7 @@ jobs:
           git config --local user.name "GitHub Action"
 
           git fetch origin gh-pages:gh-pages
-          # TODO: Remove the -f Roq 2.1.8
-          git checkout -f gh-pages
+          git checkout gh-pages
 
           rsync -a --delete --exclude='.git' --exclude='.nojekyll' "$SITE_DIR"/ .
           git add -A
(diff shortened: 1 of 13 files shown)


rolfedh:
Review in progress...

rolfedh (commented):
One thing to fix before merging, then this is good: `content/versions/2.13/guides/_attributes-local.adoc` gained a blank line where the three lines came out (it's the only file with a `+1`). That file is included from the document header via `_attributes.adoc`, a blank line there ends the header, and the `:extension-status: preview` that nine 2.13 guides set right after the include stops counting — their "preview" badge disappears. Deleting the blank line fixes it.

The revert itself is a no-op, verified with full all-versions builds: comparing the PR build against a build of its base commit, every generated page on the site is byte-identical except those nine, and with the blank line removed the count is zero. CI couldn't show this — `include_versions` came out false, so it built with `only-latest-guides` and skipped the crawler, never rendering the versioned guides.

Dev mode isn't covered by CI either; a quick `quarkus dev` xref check would close that.

gsmet (approved):

gsmet (changes requested):
See @rolfedh 's comment.

holly-cummins:
Ohh, good spot on the blank line, thanks! Fixed now. I'd also already confirmed dev mode.

gsmet (approved):

PR #2988, opened 2026-09-04
Add tests for files users might not see, but crawlers use

holly-cummins:
Slightly closing the barn door after the https://github.com/quarkusio/quarkusio.github.io/pull/2958 horse, but regressions here would be hard for users to spot.

This test did find another problem, though - sitemap.html is missing, and I don't think that's intended. I've raised https://github.com/quarkusio/quarkusio.github.io/issues/3023 and disabled the test.

### DIFF
--- a/src/test/java/io/quarkusio/DnsVerificationTest.java
+++ b/src/test/java/io/quarkusio/DnsVerificationTest.java
@@ -1,0 +1,53 @@
+package io.quarkusio;
+
+import java.io.IOException;
+import java.net.URI;
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+public class DnsVerificationTest extends BrowserTest {
+
+    @Override
+    @BeforeEach
+    void createContext() {
+    }
+
+    @Override
+    @AfterEach
+    void closeContext() {
+    }
+
+    @Test
+    void atprotoDidReturns200() throws IOException, InterruptedException {
+        try (HttpClient client = HttpClient.newHttpClient()) {
+            HttpRequest request = HttpRequest.newBuilder()
+                    .uri(URI.create(baseUrl + "/.well-known/atproto-did"))
+                    .GET()
+                    .build();
+            HttpResponse<String> response = client.send(request, HttpResponse.BodyHandlers.ofString());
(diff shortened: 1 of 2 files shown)


holly-cummins:
Interestingly, while this has been in draft, we lost llms.txt again. Not sure how that happened, but I guess it shows the value of the test.

src/test/java/io/quarkusio/DnsVerificationTest.java
```diff
@@ -0,0 +1,53 @@
...
+import java.net.URI;
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
```
marko-bekhta, line 13: 🙈 🙂 ⭐

src/test/java/io/quarkusio/SitemapTest.java
```diff
@@ -119,4 +118,78 @@ void sitemapIncludesHomePage() throws IOException, ParserConfigurationException,
...
+            HttpRequest request = HttpRequest.newBuilder()
+                    .uri(URI.create(baseUrl + "/sitemap/"))
+                    .GET()
+                    .build();
+            HttpResponse<String> response = client.send(request, HttpResponse.BodyHandlers.ofString());
+            String body = response.body();
+            assertFalse(body.isEmpty(), "sitemap should not be empty");
+            // Choose a typical url that should be in the sitemap, but not in the header menu
+            assertTrue(body.contains("<a href=\"/guides/versions"), "sitemap should contain anchor links");
+            assertTrue(body.contains("quarkus.io"), "sitemap should contain quarkus.io URLs");
```
marko-bekhta: should it be one with a protocol then ? (just wondering if this wouldn't just catch some title or header somewhere ?)

marko-bekhta (approved):

src/test/java/io/quarkusio/SitemapTest.java
holly-cummins: You're right, that would be stronger. Although actually, I don't know if the line even makes that much sense, or what exactly it's testing. It might just be catching something from the menu. But I'll update it.

src/test/java/io/quarkusio/DnsVerificationTest.java
holly-cummins: I guess we don't have impsort set up in CI. Or I don't have my IDE set up properly. Or both. Will fix!

marko-bekhta (approved):
thanks 🙂

PR #2996, opened 2026-09-08
Rewrite "To Cache or Not to Cache Virtual Threads" on measured data

franz1981:
The original post explained the throughput drop on a small heap with a mechanism that turned out to be wrong (doh! there was a Jackson bug in the benchmark used :cry: ). 
This rewrites it around what pooling virtual threads actually buys and costs, measured end to end on the same benchmark.

**What changed**

- **What pooling buys:** about 8% less allocation per request. Real in every run — and it moves nothing. Young collections, Full GCs, pause time and throughput all sit inside the run-to-run spread.
- **What it costs:** about 60 MB of live heap, roughly 6 KB per pooled thread, held for the life of the pool whether the thread is working or idle.
- **Why:** one promoted `StackChunk` per pooled thread, minted during JIT warm-up when the request path was still interpreted or C1 and so froze about four times as many stack words. A chunk's capacity is fixed when it is allocated; a promoted chunk is never written into again, so later freezes go into a small tail in front of it; and only the tail is ever thawed from, so the parent keeps the frames the JVM pushed to start the thread. A pooled worker never returns out of its loop, so those frames never leave. An unpooled thread unwinds all the way out and its chunk goes with it. A new diagram (`stackchunk-warmup.svg`) shows the split.
- **A note on a smaller heap:** at 900 MB, a third of pooled runs and no unpooled run fell into hundreds of Full GCs, with the collector logging that predicted promotion had exceeded free old-generation space. Reported as an observation, without an explanation, because we do not have one.

The conclusion is unchanged — don't pool virtual threads — but the reason is duller than the original post claimed: pooling saves the thing the collector already handles for free, and keeps the thing it does not.

Credits @pchilano, whose review of the original findings on openjdk/loom#228 is what led here.

**Notes for reviewers**

- Two files: `content/posts/2026-07-23-to-cache-or-not-to-cache-virtual-threads.adoc` and one new image, `public/assets/images/posts/to-cache-or-not-to-cache-vts/stackchunk-warmup.svg`. The three existing images are still used.
- Length is down from ~3,500 to ~1,950 words.
- Every figure comes from repeated runs on a fixed-frequency machine; the article gives ranges wherever the spread matters. All runs use ParallelGC — the cost measured here is promotion behaviour, which is collector-specific, and G1 was not tested.

### DIFF
--- a/content/posts/2026-07-23-to-cache-or-not-to-cache-virtual-threads.adoc
+++ b/content/posts/2026-07-23-to-cache-or-not-to-cache-virtual-threads.adoc
@@ -3,10 +3,28 @@ layout: post
 title: "To Cache or Not to Cache Virtual Threads"
 date: 2026-07-28
 tags: virtual-threads performance loom
-synopsis: "The OpenJDK team says don't pool virtual threads. We tested that advice with data, tried to fix the JVM when pooling broke, and discovered why the fix made things worse."
+synopsis: "The OpenJDK team says don't pool virtual threads. We measured what pooling buys - 8% less allocation - and what it costs - 60 MB of near-empty stack chunks - and found that it saves the cheap thing and keeps the expensive one."
 author: fnigro
 ---
 :imagesdir: /assets/images/posts/to-cache-or-not-to-cache-vts
+
+[IMPORTANT]
+.This article was substantially revised on 11 September 2026
+====
+The version published on 28 July 2026 explained what we measured with a cause that turned out to be wrong: the benchmark it rested on had a bug. https://github.com/pchilano[Patricio Chilano Mateo] questioned the explanation in a https://github.com/openjdk/loom/pull/228#issuecomment-5529741290[review of the original findings], and following that question is what turned the bug up. Everything below has been re-measured. The advice — don't pool virtual threads — is unchanged; the reason for it is not.
+
+.What exactly changed
+[%collapsible]
+=====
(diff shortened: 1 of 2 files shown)


geoand (approved):
Thanks!

Sanne (changes requested):
It's nice and it's great that we correct on mistakes, but I'm torn on one thing: it's usually a big No-No to update a published article w/o making it very clear that you changed it.

I think it needs some form of notice at the very top to highlight that you updated it? You mention it at the very end but it's not very visible. And having some kind of information-box which explains why you felt the need to change it and what exactly changed would be good too.

franz1981:
@Sanne you're right, and thanks — the only acknowledgement was a credit line at the bottom that didn't actually say anything had changed.

The article now opens with an `IMPORTANT` box: what was wrong (the benchmark served JSON through Jackson's default `ThreadLocal` buffer recycling, which pinned ~200 MB in the pooled arm and contaminated exactly the comparison under test), why it was revisited (@pchilano questioned the explanation in a review on openjdk/loom#228), and that the conclusion is unchanged while the reason for it is not. Inside it a collapsed "What exactly changed" lists the five differences, including what the original blamed and this version doesn't, and what was dropped — plus a permalink to the superseded version, so nothing is quietly disappeared.

I kept `date:` at the original 28 July rather than bumping it, so the correction doesn't resurface in the feed as a new post; the revision date is in the notice. Happy to change that if you'd rather.

Sanne (approved):
Very nice summary at the top! thanks

PR #3000, opened 2026-09-10
Fix missing authors page content

### LINKED ISSUE #2999: Authors page is missing authors
This got missed in the Roq conversion.https://es.quarkus.io/author/ has authors, https://quarkus.io/author/ does not. 

We maybe also are missing a test. 

We maybe would want a fancier author page, but I will leave that to another issue.

holly-cummins:
Resolves #2999 

The authors page did not list authors, despite tests asserting it did. 

This happened because of a combination of bugs:
- The test checked for a decent number of `/author/*` entries, but it so happens that the internationalisation links at the top matched this pattern. It also happened that 'decent' was 5, which is exactly how many internationalised sites we have. 
- The key for the sort needed quotes 
- The sort virtual method was missing

### DIFF
--- a/src/main/java/io/quarkus/tools/migration/JekyllFiltersExtension.java
+++ b/src/main/java/io/quarkus/tools/migration/JekyllFiltersExtension.java
@@ -178,6 +178,38 @@ public class JekyllFiltersExtension {
         return sorted;
     }
 
+    /**
+     * A simple key/value pair whose properties are named exactly as the templates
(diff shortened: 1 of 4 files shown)


marko-bekhta (approved):
makes sense 👍🏻 🙂 

> It also happened that 'decent' was 5, which is exactly how many internationalised sites we have.

ohh that's an unfortunate coincidence 😖

PR #3019, opened 2026-09-17
Fix id expansion in generated config docs to avoid unresolved `{summaryTableId}`

### LINKED ISSUE #2962: Fragment anchors generated with {summaryTableId}
This is a Roq regression. It should be expanded, and it's not.

Another one discovered by #2713.

When this is fixed, the exclusion in #2713 should be removed.

I hadn't actually dug into where these happen, but it's in every config table with a duration, for example https://quarkus.io/version/3.8/guides/hibernate-search-orm-elasticsearch/index.html:

Image

Inspecting the link shows it's Not Good:

Image

holly-cummins:
Fixes #2962 
For reasons I understand even less, it also makes most of the fix in https://github.com/quarkusio/quarkus/pull/55413 take effect (previously the apparently correct fix was a no-op). 

Generated config docs up to 3.15 look like this (later versions pre-expand the id, which is why only old versions are affected):

```
:summaryTableId: quarkus-agroal_quarkus-datasource [.configuration-legend]
icon:lock[...] Configuration property fixed at build time - ... [.configuration-reference.searchable, cols="80,.^10,.^10"] |
===
... |Duration link:#duration-note-anchor-{summaryTableId}[...] 
```

Because that file is `include`d in a guide, `summaryTableId` is a body attribute entry. Asciidoctor calls `restore_attributes` at the end of parsing, so body attributes only exist during conversion via playback: the entry is attached to the next block (here the legend paragraph) and replayed when that block is converted. Table cells apply their substitutions at convert time, so they depend on that playback.

`ConfigTableTreeprocessor` was replacing the legend paragraph with a brand-newblock `createBlock` to inject search input. The replacement dropped the attribute entries, so the playback never happened and every table cell rendered the literal `{summaryTableId}`. The   `[id='duration-note-anchochor]`, resolved at parsetime, expanded correctly. Hence links pointing at ids that don't match.
                                                                            
A second, silent symptom was that `configuration-legendrole` was also being dropped.

The fix is that to append the search input to the existing legend block via `setSource()`.

I confess my understanding here is limited, and Claude did rather a lot of heavy lifting. :)


src/main/java/io/quarkus/tools/migration/asciidoc/ConfigTableTreeprocessor.java
```diff
@@ -59,16 +59,14 @@ public Document process(Document document) {
...
-                    Block newCaption = createBlock(parent, "paragraph",
-                            captionContent + " <input type=\"search\" id=\"config-search-%d\" placeholder=\"FILTER CONFIGURATION\" disabled>".formatted(searchFieldId),
-                            new HashMap<>());
-                    siblings.set(tableIndex - 1, newCaption);
+                    // Append to the legend block rather than swapping in a fresh one. A
+                    // replacement block loses the caption's roles, and it loses the attribute
+                    // entries Asciidoctor replays when converting it
+                    caption.setSource(caption.getSource()
+                            + " +++<input type=\"search\" id=\"config-search-%d\" placeholder=\"FILTER CONFIGURATION\" disabled>+++"
+                                    .formatted(searchFieldId));
```
marko-bekhta, line 69: I'm a bit puzzled 😟 🙈 🙂 we are changing how we add the filter input block for config properties, but I clicked around on the site (both guides and all props) for different versions and things seem fine, so I'm missing what we were trying to address 🙈

holly-cummins: Ah, let me find a page with the problem. It's subtle, just dead links.  And dead fragment links, so they look right until you look really closely.

holly-cummins: I added the details in https://github.com/quarkusio/quarkusio.github.io/issues/2962#issue-5279230243. I didn't really know them until you asked, because I'd just had a red line in a test. It's all the little black question marks next to durations in config tables. We're making links like https://quarkus.io/version/3.8/guides/hibernate-search-orm-elasticsearch/index.html#duration-note-anchor-{summaryTableId}. I think the problem got fixed after 3.15, I guess because we changed how we did things, so it's sort of hard to validate in the preview, which doesn't have versions which show the problem. But I did check that /main/ in the preview has well-formed Duration links.

marko-bekhta (approved):
ok, that 💎 took me a moment 🙈 🙂 but at least now I somewhat understand why the fix should work 🙂 

btw, unrelated to this PR -- I think we do a double-registration of extensions in the tests, they seem to also pick up the service file, so extension might run twice on a single file ? 🫣

holly-cummins:
> ok, that 💎 took me a moment 🙈 🙂 but at least now I somewhat understand why the fix should work 🙂

Thank you! 

> btw, unrelated to this PR -- I think we do a double-registration of extensions in the tests, they seem to also pick up the service file, so extension might run twice on a single file ? 🫣

Oh dear. Would https://github.com/quarkusio/quarkusio.github.io/issues/3028 help? Probably not. Can you raise an issue?

PR #3022, opened 2026-09-17
Expect s2i fragment to work in newer versions

holly-cummins:
Will fail until https://github.com/quarkusio/quarkus/pull/56837 merges and syncs; raising it so we don't forget to strengthen the test.


holly-cummins:
Weirdly, https://github.com/quarkusio/quarkus/pull/56837 fixed one broken reference, but not the other.
Oh! It's because it's checking what's in the released version as well. I'll see if we can backport the fix.

### Before:

Broken fragment anchors:

    #s2i http://localhost:8090/guides/container-image ← http://localhost:8090/guides/deploying-to-kubernetes
    #s2i http://localhost:8090/version/main/guides/container-image ← http://localhost:8090/version/main/guides/deploying-to-openshift/

### After 

Broken fragment anchors:

    #s2i http://localhost:8090/guides/container-image ← http://localhost:8090/guides/deploying-to-kubernetes

holly-cummins:
3.40 is out, so this is good.

PR #3032, opened 2026-09-18
Restore guide redirects lost in the move to Roq

rolfedh:
The move from Jekyll to Roq did not carry over 21 guide redirects, so their old URLs return 404, and it turned one more, `security-openid-connect-web-authentication`, into an empty page. Released Quarkus versions still link to some of these from extension metadata, for example `/guides/security-openid-connect` for `quarkus-oidc`. This restores all 22 with their original targets and adds a test for each. `opentelemetry` and `security-webauthn` stay gone, because both are live guide names again. It also removes the six `LinkCrawlerTest` suppressions for these URLs, so the crawler guards them again.


rolfedh:
@holly-cummins could you take a look at this one? It restores the 21 guide redirects that the Roq conversion folded into the guide frontmatter as `aliases:`, which the next release sync then overwrote. The detail is in your thread on quarkusio/quarkus#56875.

I also pushed a second commit removing six entries from `KNOWN_FAILURES` in `LinkCrawlerTest`, the ones under the "worked on Jekyll but not with Roq" TODO. All six answer 200 on the preview and refresh to the guide they name, so the crawler should stay green.

holly-cummins:
Will do! Although in principle I'd prefer to get the Roq functionality redirect capability doing what we need, rather than work around it. 

I guess in this case we never thought Roq was wrong, we just accepted that the redirect syntax was (quite reasonably) different between the two. And then we attempted to map what was in source control, but , we (ie the converter) just made a change to the guides files on quarkusio. Such a change was misguided, and of course would then get lost on the next sync, as you say, regressing links that had appeared to work post-conversion. 

So the real failure was not a Roq problem, but failing to notice the impermissible guides metadata change amongst the other 40,000 changes. I *did* ask Claude to look for guides changes, but I think even Claude was a bit overwhelmed by the volume. So as I typed I've talked myself into thinking that this is absolutely the correct fix. :)

holly-cummins:
I think another thing that happened is that I decided I preferred having the aliases in the metadata to the separate redirect file, because it was neater. The redirect file has both the advantage and disadvantage of being an overlay in a separate repo. It means it's a bit of extra clutter, and changes might need to happen in two places, but it means we get better control over the lifecycle and it's not chained to the support lifecycle.

holly-cummins:
@rolfedh, do you think we should handle any of these as `aliases` in the quarkus repo? There would be a couple of advantages of doing it over there; reduced clutter of files, and also, having the new syntax in the guide content, so it's there for people to learn from. But there might also be some disadadvantages, too (and we'd need to wait for the release cycle to complete, possibly several cycles, before it would take effect).

rolfedh:
Thanks, and the credit for spotting the cause goes to you. Twenty-one alias lines among forty thousand changed files was never going to show up in review; it only turned up by diffing the guide files either side of the sync.

rolfedh:
Your reasons for the quarkus repo are good ones, and I think the answer splits: not these 22, but yes for renames from here on.

These 22 need to work now. 20 of them are 404 today, and the other two land on the empty page this PR deletes. `content/guides` is only rewritten when the sync runs for a release, so an alias would not take effect until then.

For a rename, the quarkus repo is the better home. Front matter there survives the sync (`all-config.adoc` keeps its `layout:` all the way into `content/guides`), and the alias travels with the rename in one pull request, which is exactly the example you want people to copy.

One thing to weigh either way: an alias lives and dies with its guide file. Rename or retire that guide later and the alias goes with it, quietly. A redirect file outlives it.

holly-cummins:
Ok, let's merge these now and maybe do a follow-on piece of work to add some of these aliases back on the Quarkus side for posterity?

holly-cummins (approved):

rolfedh:
Thanks, @holly-cummins. I'll return and create a PR with the follow-on work later this week.

rolfedh:
@holly-cummins 

Before I open that pull request, something I should flag: adding the aliases on their own would break the 22 redirects we just merged.

The nightly sync writes `content/versions/main/guides`, and only a release sync writes `content/guides`. An alias in a versioned copy still claims the site root path, and it redirects to the page that declares it. So the night after the quarkus pull request merged, `/guides/resteasy-reactive` would point at `/version/main/guides/rest/` instead of `/guides/rest`, shadowing the redirect file. I built the site four ways locally to check. The build passes with no warning, `RedirectTest` fails with exactly that, and when both copies carry the alias the versioned one still wins, so a release does not clear it. This is not specific to these 22. It would happen for any rename that follows the instruction in `doc-contribute-docs-howto.adoc`.

The fix looks small: a post-processing step in `sync-web-site.sh` that rewrites `/guides/X` to `/version/${BRANCH}/guides/X` when the target is a versioned directory. With that applied, the root path went back to `/guides/rest` and the snapshot got its own working redirect. One more detail: the alias emits a trailing slash, so `RedirectTest` needs to accept both forms.

So it is two small fixes plus the aliases, across both repos, rather than one pull request. Happy to do all of it. Is that the tradeoff you want, or would you rather leave the redirect files as they are?

rolfedh:
Before I build that sync step, would you rather I fixed this in Roq instead? I can, but each upstream option I can see has a real cost, so I would rather you chose.

I ask because of your earlier point: "in principle I'd prefer to get the Roq functionality redirect capability doing what we need, rather than work around it." The sync step is a workaround by that standard.

Making an alias resolve relative to its own collection would fix it at the root, but it would break the 72 `.html` files in `content/redirects/guides`, which work only because their aliases resolve from the site root. Adding detection is smaller and safe: Roq throws when a page collides with a page, but an alias colliding with a page is checked nowhere, which is why my build stayed silent while the alias quietly won. That would make the next occurrence loud, though on its own it would not make a versioned alias work.

rolfedh:
I went ahead with the smaller of the two Roq options and opened quarkiverse/quarkus-roq#1253, since a missing collision check stands on its own whichever way you answer the rest. One thing that came out of testing it: an alias can silently delete a live guide, not just shadow a redirect file.

